PR #493, opened 2020-11-28
Bricktheworld/develop

SoPat712:
Puts scroll to bottom indicator for material theme in the bottom right as a circle, like this 
image

Delete the extra files pls. Also you spelled attachment wrong so I think I changed that.


Bricktheworld:
I will manually merge these

SoPat712:
Great, thanks

SoPat712:
Also idk if it's just a placebo or something but I feel like scrolling is less laggy with the indicator always on the bottom right.

PR #833, opened 2021-05-03
Fixed issue with setting not actually saving

jjoelj:
Previously I was not aware of the additional necessary code to save a new setting. Lmk if there are any other places I missed. The setting actually saves now.


zlshames (approved):
Probably need a little better of a system for this so we don't have to make so many changes to add a config item. However, for now, we'll continue using this method

PR #901, opened 2021-06-19
Add submenu UI for message details popup

### LINKED ISSUE #898: Create another menu within message details popup for additional context options
Currently, the message details popup (when you tap and hold a message), provides a context menu for a few actions. Well, we want to add additional actions such as "forward", "start conversation", etc. The only way to do that is to create a different menu with those items. We probably also want to move some of the attachment download options to that other menu as well. This way the context menu stays de-cluttered.

This is marked as medium because it involves UX and deciding the best place to put that menu. I'm leaning towards a floating action menu button that sits either on the left or right side of the screen, depending on which message was selected. When tapped, show the other options.

tneotia:
Fixes #898.

**Note** Only the UI has been added here, the "internals" still need to be hooked up, i.e. the DM/forwarding functionality needs to be implemented. I don't have enough knowledge of the library to successfully implement these yet.

| Screenshot 1 | Screenshot 2 |
| - | - | 
| image | image |

### DIFF
--- a/lib/layouts/widgets/CustomCupertinoAlertDialog.dart
+++ b/lib/layouts/widgets/CustomCupertinoAlertDialog.dart
@@ -1,0 +1,1788 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// @dart = 2.8
+
+import 'dart:math' as math;
+import 'dart:ui' show ImageFilter;
+
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/foundation.dart';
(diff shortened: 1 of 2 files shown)


zlshames:
Overall great work. I want to merge this with Joel's implementation of forwarding and replying. Do you think you could make the popup a little bit more dynamically skinned based on the being material vs iOS? The boxy-ness looks too material for Cupertino

tneotia:
Fixed this, now it looks like:

| iOS | Material | 
| - | - | 
| image | image |

Unfortunately I had to add a custom `CupertinoAlertDialog` because Flutter doesn't allow you to customize the background color....

zlshames (commented):
What custom changes did you make in CustomCupertinoAlertDialog?

PR #904, opened 2021-06-20
Support for group chat naming

### LINKED ISSUE #900: When a group chat doesn't have a title, don't show the "NAME" field in details

tneotia:
Fixes #900.

Group chat naming only works locally for now.

### DIFF
--- a/lib/layouts/conversation_details/conversation_details.dart
+++ b/lib/layouts/conversation_details/conversation_details.dart
@@ -37,6 +37,7 @@ class _ConversationDetailsState extends State<ConversationDetails> {
   bool isCleared = false;
   int maxPageSize = 5;
   bool showMore = false;
+  bool showNameField = false;
 
   bool get shouldShowMore {
     return chat.participants.length > maxPageSize;
@@ -55,6 +56,7 @@ class _ConversationDetailsState extends State<ConversationDetails> {
     super.initState();
     chat = widget.chat;
     controller = new TextEditingController(text: chat.displayName);
+    showNameField = chat.displayName.isNotEmpty;
 
     fetchAttachments();
     ChatBloc().chatStream.listen((event) async {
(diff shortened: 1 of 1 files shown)


lib/layouts/conversation_details/conversation_details.dart
```diff
@@ -128,21 +130,81 @@ class _ConversationDetailsState extends State<ConversationDetails> {
...
+                      children: [
+                        Expanded(
+                          child: Padding(
+                            padding: const EdgeInsets.only(left: 16.0, right: 8.0, bottom: 8.0),
+                            child: TextField(
+                              cursorColor: Theme.of(context).primaryColor,
+                              readOnly: !chat.isGroup(),
+                              onSubmitted: (String newName) async {
+                                await widget.chat.changeName(newName);
+                                widget.chat.getTitle();
```
zlshames: This probably should be awaited, right? Or did you purposefully not await it?

lib/layouts/conversation_details/conversation_details.dart
```diff
@@ -128,21 +130,81 @@ class _ConversationDetailsState extends State<ConversationDetails> {
...
+                              autocorrect: false,
+                              decoration: InputDecoration(
+                                labelText: chat.displayName.isEmpty ? "SET NAME" : "NAME",
+                                labelStyle: TextStyle(color: Theme.of(context).primaryColor),
+                                enabledBorder: UnderlineInputBorder(borderSide: BorderSide(color: Colors.grey))
+                              ),
+                            ),
+                          ),
                         ),
+                        if (chat.displayName.isEmpty)
```
zlshames, line 162: Should you use the variable you created above here?

lib/layouts/conversation_details/conversation_details.dart
zlshames: I'm dumb sorry. Didn't look at your next commit

zlshames:
Also, I'm wondering if the group title will be overwritten when a new message comes in from that chat, and the .update function gets called in the chat model. Cuz that is supposed to set the group title if it's different from what we have in the DB. 

Should we add a DB column in the chat table for if the chat has a custom title set? And then in the chat details page, we have a reset button if that is true?

lib/layouts/conversation_details/conversation_details.dart
tneotia: Not here because we only want to show the "cancel" button next to the TextField if the displayName is empty (i.e. the user is adding a name so they should be able to cancel that action).

tneotia:
My group chats are very inactive so I can't really test that lol... But just to clarify, you're saying that when a new chat comes in, the app will update the name back to what it is in the server? If it just pulls from the DB it shouldn't change since I update both displayName and title.

A reset button might be a good idea, but at the same time I don't know if that makes sense UX-wise because it might be confusing to the user, like what exactly am I resetting it back to? Is it going to reset the name for everyone? Etc.

zlshames:
> My group chats are very inactive so I can't really test that lol... But just to clarify, you're saying that when a new chat comes in, the app will update the name back to what it is in the server? If it just pulls from the DB it shouldn't change since I update both displayName and title.
> 
> A reset button might be a good idea, but at the same time I don't know if that makes sense UX-wise because it might be confusing to the user, like what exactly am I resetting it back to? Is it going to reset the name for everyone? Etc.

I think it might be ok actually. Looking at the code here: https://github.com/BlueBubblesApp/BlueBubbles-Android-App/blob/master/lib/repository/models/chat.dart#L263-L265

It looks like we only update the displayName if the current displayName is null. So if you give it a name when it's empty, it won't get updated. I think this is good then.

PR #913, opened 2021-06-22
Fix contact avatar showing the other person instead of the one who actually sent the tapback

### LINKED ISSUE #908: If you're in an individual chat, and the person you're chatting with has a contact photo, the "You" avatar on tapback menu has the recipient's contact photo
I know the title is a mouthful, so feel free to abridge it. This is somewhat related to issue #763, but it's not exactly the same kind of bug. Below is the contact information for an individual, as well as the tapback menu in a conversation that I had with that same person. The avatar displays the recipient's contact photo instead of a blank one or my own (would it be possible to display my own contact photo?)

This is not an issue in group chats.
Screenshot_20210620-212941_Contacts
Screenshot_20210620-213003

tneotia:
Fixes #908.

Tried to get "my" contact photo in there but I didn't see a way to do this, especially because in a group chat "my" handle does not come up in the list of participants, making me think that "my" contact details aren't sent from the iMac? Not too sure but for now it will display a "Y" as the avatar.


zlshames (approved):
👍🏽

PR #915, opened 2021-06-22
Add support for .heic images in fullscreen mode and show a proper error when the image fails to load

### LINKED ISSUE #911: show unsupported extension error rather than broken picture placeholder in image viewer
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@jojejo` in `#support`.

tneotia:
Fixes #911.

Not the best solution because .heic will still appear slightly compressed in the fullscreen viewer but its better than not being able to see the image at all.

### DIFF
--- a/lib/layouts/image_viewer/image_viewer.dart
+++ b/lib/layouts/image_viewer/image_viewer.dart
@@ -72,10 +72,11 @@ class _ImageViewerState extends State<ImageViewer> with AutomaticKeepAliveClient
   }
 
   Future<void> initBytes() async {
-    bytes = await widget.file.readAsBytes();
     if (widget.attachment.mimeType == "image/heic") {
       bytes = await FlutterImageCompress.compressWithFile(widget.file.absolute.path,
           quality: 100);
+    } else {
+      bytes = await widget.file.readAsBytes();
     }
     if (this.mounted) setState(() {});
   }
@@ -279,7 +280,7 @@ class _ImageViewerState extends State<ImageViewer> with AutomaticKeepAliveClient
                         return loader;
                       },
                       loadFailedChild: Center(
-                          child: Text("Failed to load image",
+                          child: Text("Failed to display image",
                               style: TextStyle(fontSize: 16, color: Colors.white))
                       )
                     )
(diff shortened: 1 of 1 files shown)


jjoelj (changes requested):
* Change error text to "Failed to display image" rather than "Failed to load image"
* Use an if-else in `initBytes()` Don't need to use flutter's `readAsBytes()` for no reason
* make the text style of the error widget theme-reliant using a property of `Theme.of(context)`

tneotia:
> * make the text style of the error widget theme-reliant using a property of `Theme.of(context)`

@jjoelj for that last one the viewer is black in light mode too so we don't want black text on black BG using `Theme.of(context)`. Is there some other property that I can use? 

Technically I could check if the color is black and correct it but it's not ideal.

jjoelj (approved):
> > * make the text style of the error widget theme-reliant using a property of `Theme.of(context)`
> 
> @jjoelj for that last one the viewer is black in light mode too so we don't want black text on black BG using `Theme.of(context)`. Is there some other property that I can use?
> 
> Technically I could check if the color is black and correct it but it's not ideal.

Now that I look at the existing code, it looks like a lot of the colors are hardcoded. In that case this looks good. We can open a separate ticket for making the photoviewer as a whole more responsive to the theme.

zlshames (approved):
Agreed, not a long term solution, but a step in the right direction

lib/layouts/image_viewer/image_viewer.dart
```diff
@@ -273,6 +279,10 @@ class _ImageViewerState extends State<ImageViewer> with AutomaticKeepAliveClient
                       loadingBuilder: (BuildContext context, ImageChunkEvent ev) {
                         return loader;
                       },
+                      loadFailedChild: Center(
```
zlshames, line 282: Before I merge this, can you share what this widget will look like? Is it consistent with what we have in other areas?

zlshames:
There is another ticket regarding .heic support. Since this is not a full solution, I am leaving that one open

Issue #253, opened 2020-09-25
Ability to fetch and save metadata around the server (OS version)


zlshames:
This is so we can determine if there are some features that we shouldn't show on the app

zlshames:
This should be fairly easy to do. We should have a page that you can go to that will fetch the info from the server. The endpoint already exists.

PR #928, opened 2021-06-23
Support showing server metadata to users

tneotia:
Don't know if this is what #253 is after, but I think its useful anyways and can pave the way to add more server info accessible from the app.

image


zlshames (approved):

zlshames:
Yup, this is more or less what i meant

Issue #933, opened 2021-06-24
[Material] allow deleting chats from archive


cynical89:
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@jojejo` in `#community-contributions`.

Issue #932, opened 2021-06-24
[Material] make muted indicator blue if muted chat has unread message


cynical89:
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@jojejo` in `#community-contributions`.

Issue #931, opened 2021-06-24
[Material] make pinning, mark as read, and mute/unmute options available if all selected chats have the same values for each of those


cynical89:
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@jojejo` in `#community-contributions`.

Issue #925, opened 2021-06-23
Add ability to mark chats as unread from chat list in material theme


cynical89:
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@jojejo` in `#android-bugs`.

zlshames:
@jjoelj has said he will pick this up

PR #935, opened 2021-06-24
Joel/features/improved conversation list

### LINKED ISSUE #815: Add typing indicators to chat list
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@zach` in `#support`.

### LINKED ISSUE #924: Fix asymmetrical padding on conversation list
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@tneot01` in `#android-bugs`.

jjoelj:
### Summary
Fixes #815, #924, #925, #931, #932, #933

* Also makes several icons more consistent across screens and themes as I saw fit
* Unpinning icon is now an empty star

### Specifics
* #815
  * Pretty self explanatory
* #924
  * Increased padding on the right arrow and decrease padding on the left of the listtiles. Decreased pin/mute/unread indicators on ios to accommodate the change.
* #925
  * Added unread/read button to the action buttons when chats are selected
* #931
  * Made all bulk action buttons depend on the selected chats
    * If all selected chats can be pinned, the pin button will show up
    * If all can be unpinned, the unpin button will show up
    * etc.
    * The icons themselves also change based on what action (positive or negative) they are going to do
  * The archive button is constant on the main chat list
  * The unarchive and trash buttons are constant on the archived chat list
* #932
  * This one's pretty self explanatory
* #933
  * Added a trash action

### DIFF
--- a/lib/layouts/conversation_list/conversation_list.dart
+++ b/lib/layouts/conversation_list/conversation_list.dart
@@ -858,7 +858,27 @@ class __MaterialState extends State<_Material> {
                         Row(
                           mainAxisAlignment: MainAxisAlignment.end,
                           children: [
-                            if (selected.length <= 1)
+                            if (([0, selected.length])
+                                .contains(selected.where((element) => element.hasUnreadMessage).length))
+                              GestureDetector(
+                                onTap: () {
+                                  selected.forEach((element) async {
+                                    element.hasUnreadMessage = !element.hasUnreadMessage;
+                                    await element.save(updateLocalVals: true);
+                                  });
+                                  if (this.mounted) setState(() {});
+                                  selected = [];
+                                  setState(() {});
+                                },
+                                child: Padding(
+                                  padding: const EdgeInsets.all(8.0),
+                                  child: Icon(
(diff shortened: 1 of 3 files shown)


zlshames:
@jjoelj Hey, so I see one issue with the typing indicators. Right now, in `CurrentChat`, we have a stream for what I can only assume is a generic, "hey, the current chat updated in some way". I wasn't the one to set this up, which is why I'm only assuming (though, I'm confident). What I think we need to change it to instead, is a more stateful stream.

For instance, I think within `current_chat.dart` we need to create an `enum` for `CurrentChatEvent`, and then make the stream handle a `Map<CurrentChatEvent, dynamic>`. That way, when listening on it, we know what type of update event it is, and we can handle it accordingly. Right now, it only handles a few event types, but without the EventType check, we might be updating widgets that we don't need to be updating.

I also think that once this is done, we can actually use the stream within a `StreamBuilder` instead of using a state variable within ConversationTile. So instead of updating the entire tile when there is a typing indicator, we only change the stream builder.

Make sense?

jjoelj:
I think I made all the changes you requested. Lmk if I did everything in an acceptable way.

zlshames (approved):
Yup, exactly.

PR #949, opened 2021-06-24
Switch from flutter_socket_io to socket_io_client

tneotia:
Title.

The purpose is to help remove custom dependencies and prepare for nullsafety migration with regularly-maintained, well-liked deps.

This will need some *extensive* regression testing, and potentially an alpha release.


zlshames (commented):
Hey, before I approve this. Did you make sure to test socket connections for when the app is fully closed? For instance, when the app is closed, you receive a notification, and you reply directly from the notifications shade. Does the message still go through?

I assume the answer is yes since you mostly didn’t change that logic, but I want to make sure nothing in the socket plugin broke.

Also, side question, did you see any performance improvements by any chance?

tneotia:
Ah shoot I didn't test any notification stuff. Honestly it might have broken because I would assume it requires platform specific code? Don't know for sure.

Right now my app is on the getx branch so it doesn't have these changes, would you be able to test this stuff out on your end? If not I can test it once I'm back home but someone will have to text me lol.

I am just connected via IP so on my end - I saw latency between 30-100ms. Don't know if that's better or worse though.

zlshames:
There's no rush for this, so we can just wait till you got some time. Also, I took your getx changes, made my own finishing touches, then pushed it to development, so that should be good to go to use for stable development

tneotia:
@zlshames ready to go! Notifications come through and I can reply through the notification panel just fine. Super happy with how this one went.

zlshames:
Great job!

PR #950, opened 2021-06-24
Disable creating new chats if the server is on Big Sur or up

### LINKED ISSUE #929: Use get-server-metadata socket endpoint to restrict creating chats on MacOS 11+
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@tneot01` in `#support`.

tneotia:
Fixes #929.

Applies opacity to FAB on conversation list, and disables pressing the "start conversation" and "forward" items in the menu options.

"Forward" is disabled because we can't (atm) prevent users from sending forwarded messages to people they don't already have a chat with.

### DIFF
--- a/lib/layouts/conversation_list/conversation_list.dart
+++ b/lib/layouts/conversation_list/conversation_list.dart
@@ -228,16 +228,21 @@ class _ConversationListState extends State<ConversationList> {
     ];
   }
 
-  void openNewChatCreator() {
-    Navigator.of(context).push(
-      CupertinoPageRoute(
-        builder: (BuildContext context) {
-          return ConversationView(
-            isCreator: true,
-          );
-        },
-      ),
-    );
+  void openNewChatCreator() async {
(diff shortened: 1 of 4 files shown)


zlshames (commented):
Hey, so one thing I’m concerned about this PR is that not _everything_ is broken on Big Sur. For instance, _sometimes_ chat creation works on Big Sur, as long as the chat being created is just with 1 person, and all the conditions fit perfectly.

I'm hesitant to make this change. I would much rather have a warning instead of it flat out being disabled. For instance, a warning popup that says, "Hey, we've detected you're using Big Sur. Chat creation is limited on this operating system due to limitations imposed by Apple"

tneotia:
Ah I was under the impression it didn't work at all. I'll rework this then and add a snack bar when entering the chat creator or something.

tneotia:
@zlshames made the changes, this should be ready to go.

zlshames (approved):

PR #954, opened 2021-06-25
Remove GetX theming implementation for now

tneotia:
Title. Will come back to this once GetX migration is complete.


tneotia:
Wrong branch oopsie

PR #964, opened 2021-06-25
Added details menu share item

### LINKED ISSUE #962: add share option to details menu
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@jojejo` in `#community-contributions`.

jjoelj:
Resolves #962 

* #962 
  * Added share option to all messages in the details menu
  * If they are attachments, `Share.file()` is used
  * If they are plain text, I added a new java method and it is called in dart using `Share.text()`


tneotia (commented):
Think you may have complicated it a bit because we could have just used https://pub.dev/packages/share_plus, but nevertheless nice work, this should be fine!

Edit: Seems like we already used a platform-specific implementation for sharing files so what I said matters even less

jjoelj:
> Think you may have complicated it a bit because we could have just used https://pub.dev/packages/share_plus, but nevertheless nice work, this should be fine!
> 
> Edit: Seems like we already used a platform-specific implementation for sharing files so what I said matters even less

Yeah I looked at share_plus but wanted to keep it consistent with our other sharing methods. I can change them all if @zlshames wants.

zlshames:
@jjoelj can you resolve these conflicts?

Also, I made a ticket for tracking migrating to share_plus at some point since they are maintaining it better than we are with our android-specific code

zlshames (changes requested):
Fix conflicts por favor :)

Issue #943, opened 2021-06-24
[GetX Migration] Migrate page navigation


cynical89:
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@zach` in `#community-contributions`.

tneotia:
We should use named routes and pass data via parameters or arguments when navigating (or create a shared Getx controller for the data if it's too cumbersome for those methods).

zlshames:
> We should use named routes and pass data via parameters or arguments when navigating (or create a shared Getx controller for the data if it's too cumbersome for those methods).

Don't we kinda already do this without GetX? What's the benefit of using GetX here?

tneotia:
Named routes are cleaner and you can do stuff like `Get.toNamed('/route?chat_title="test"')` and grab "test" in the page associated with /route. Just makes it easier to send data back and forth rather than creating a class instance for the page with the data imho. But our current way is fine.

Even if we don't want to transition fully to GetX navigation I would recommend switching to contextless navigation to make it a lot easier.

tneotia:
This is complete with the tablet mode navigation, we use GetX to push pages to the stack

PR #977, opened 2021-06-27
[WIP] Migrate all settings screens to GetX

### LINKED ISSUE #944: [GetX Migration] Migrate settings
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@zach` in `#community-contributions`.

tneotia:
Fixes #944 and will partially complete #943.

Here's the details on how the migration is done, if anyone would like to help.

1. Convert the runtime types for any settings in the screen you are migrating to `Rx`. If they have a default value, add `.obs` to the end. E.g. `int example = 10;` becomes `RxInt example = 10.obs;`.
2. `Rx` variables cannot be set directly - you cannot do `example = 15;`. You have to use the `.value` setter, so you can do `example.value = 15;`.
3. With that in mind, make sure to change `factory Settings.fromConfigEntries` and `List<ConfigEntry> toEntries()` methods to use the `.value` getter as we cannot store `Rx` directly in the DB and we cannot set `Rx` vars with regular type values.
4. Next, wrap each settings options in an `Obx(() => setting widget goes here!),`, this will ensure the `StatelessWidget` updates itself. 
5. Once you've done all that, then you can migrate the screen itself to a `GetView`. Create a `GetxController` and a `Bindings` class (see `attachment_panel.dart`) and move all vars/`initState`/`dispose` etc methods there. If any vars use `setState()` in the screens, convert them to `Rx` and use `Obx` on any widgets that use those vars.
6. Then you need to go to `main.dart` and add the page as a `GetPage` under `GetMaterialApp`. Apply your binding and give it a route name, just like I did for `AttachmentPanel`.
7. Finally, go to the screen that opens the screen you're migrating and change the `Navigator.of(context) blah blah` method to simply `Get.toNamed(route name here)`.

You may have some errors as the settings options are used in other screens, most can be resolved by putting `.value` at the end.


zlshames (commented):
This might be a dumb question, but I saw you removed all the `saveSettings()` calls. When you toggle a switch, we need it to make the DB change as well as changing the settings object's variables. Does it do that? How does it save it in the DB when a toggle is switched?

tneotia:
Afaik there was no `saveSettings` when the switch was changed, it only saved when the user exited the page. That call is still there under the `GetxCobtroller` dispose event.

If you want we can make it more reactive and save settings every time a switch/slider is changed.

zlshames:
> Afaik there was no `saveSettings` when the switch was changed, it only saved when the user exited the page. That call is still there under the `GetxCobtroller` dispose event.
> 
> If you want we can make it more reactive and save settings every time a switch/slider is changed.

Can we make it so when a toggle is changed, it saves? We want to maintain the fact that you can toggle a switch and then quit the app, then expect it still to be changed. Or will the dispose event be called when the app is quit, and it will save?

tneotia:
> Or will the dispose event be called when the app is quit, and it will save?

Don't think that's how it works because a hot restart will not save the settings unless you back out of the screen first. Should be a pretty minor thing to change then.

zlshames:
Is this the same as the other one?

tneotia:
Yeah it can be closed I got whatever I needed from it.

PR #1091, opened 2021-07-16
Add new conversation settings panel

tneotia:
Zach don't merge yet please, need to fix one small thing and then it'll be ready :)


tneotia:
@zlshames this is ready to go now

PR #1125, opened 2021-07-20
Fix colorful avatars not updating after toggled

### LINKED ISSUE #1108: colorful avatars don't take effect immediately on the conversation list after enabling the setting
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@tneot01` in `#android-bugs`.

tneotia:
Fixes #1108

### DIFF
--- a/lib/layouts/widgets/contact_avatar_widget.dart
+++ b/lib/layouts/widgets/contact_avatar_widget.dart
@@ -186,13 +186,6 @@ class _ContactAvatarWidgetState extends State<ContactAvatarWidget> with Automati
   Widget build(BuildContext context) {
     super.build(context);
 
-    Color? color1 = colors.length > 0 ? colors[0] : null;
-    Color? color2 = colors.length > 0 ? colors[1] : null;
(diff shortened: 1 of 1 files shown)


zlshames (approved):
I love that GetX allows us to better react to these config changes. It was a pain dealing with it before without a controller/state management system

zlshames:
We may have to do the same for the colorful bubbles option

tneotia:
Nah that works fine because the colorful bubbles state will never be modified while that screen is open. Colorful avatars will be since the conversation list is active while in settings.

zlshames:
> Nah that works fine because the colorful bubbles state will never be modified while that screen is open. Colorful avatars will be since the conversation list is active while in settings.

There is a super old bug that Monica originally ran into where if you turn on colorful bubbles, then go into a chat, the colors of the bubbles changed, but any new messages that come in will be the old color. 

Maybe this has been fixed throughout time and all the changes 🤷🏽‍♂️

tneotia:
Ah okay I haven't seen this personally but it might be a thing 🤷‍♂️

PR #1137, opened 2021-07-21
Potentially fix dupe messages on unstable connections

### LINKED ISSUE #612: Fix dupe sending issue
Idk why but some messages are duping when they send. Not sure if I need to implement something on the server side or on the android side. Maybe I'll just check the queue service?

### LINKED ISSUE #1070: Dupe messages when disconnect happens as message is sent
Marking this as high priority because it was not as much of an issue in the pre-flutter2 builds. Now, we are seeing dupes get sent 1, 2, 3, or even 4 times. 

We may need a caching system. I have already switched to hashing the message vars to generate a tempGUID. Probably is worth caching a list of those and checking against it. 

HOWEVER, the issue might be with the flutter socket IO library. It might detect the disconnect and assume the packet wasn't sent, even if it is. So, this might be something I need to fix on the server.

tneotia:
May fix #1070, may fix #612


lib/action_handler.dart
```diff
@@ -107,20 +108,35 @@ class ActionHandler {
...
 
-      completer.complete();
-    });
+        completer.complete();
+      });
+    };
+
+    bool isConnected = await InternetConnectionChecker().hasConnection;
+    if (!isConnected) {
+      StreamSubscription? sub;
```
zlshames, line 130: Should we add a timeout to this to cancel the subscription? I assume Apple does this as well. For instance, if it's been 1 minute and still no connection, cancel the message.

zlshames: Also, what if the user send the message (with no connection). Then leaves the app. If they come back 5 min later, will the message be timed-out? Or will it be "greyed out"? The correct answer should be that the message has an error

zlshames: Maybe we need a lifecycle checker for when the app is resumed, check for queued outgoing messages

zlshames (approved):

zlshames:
I'm leaving that second linked ticket open until we confirm that dupes have been mitigated

PR #1171, opened 2021-07-24
Improve contact matching by retaining only 0-9 when matching

### LINKED ISSUE #1058: When matching address to contact, sanitize/slug both sides
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@zach` in `#support`.

tneotia:
Fixes #1058


zlshames (changes requested):
Hey, just in case, can we can numericOnly on the lastDigits string? I don't think it should ever have any other characters, but ya know, just in case

zlshames (approved):

PR #1199, opened 2021-07-26
Potentially fix share menu performance

### LINKED ISSUE #1106: Compress images on download instead of on load
This is likely due to the attachment metadata being loaded into BB. Not sure why it lags the UI since it should all happen in the background. But maybe this is something we need to do via a dart VM so it doesn't lag the UI thread? Not sure

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@zach` in `#android-bugs`.

tneotia:
Might fix #1106, will def need confirmation


zlshames (approved):

zlshames:
Merged, but leaving the ticket open. I didn't think the share menu lagged that much, but we'll see

PR #1206, opened 2021-07-27
Remove streams created in ChatBloc

tneotia:


zlshames:
After some real world usage, it's working pretty well, and a step in the right direction. Merging

PR #1226, opened 2021-07-31
Remove keyboard attachable plugin bc it is causing exceptions

tneotia:


tneotia:
@zlshames now that we've narrowed down the keyboard issue I think this is okay to merge now as well.

PR #1229, opened 2021-08-01
better pinned chats!

jjoelj:
Add big pin design to ios theme.

### DIFF
--- a/lib/blocs/chat_bloc.dart
+++ b/lib/blocs/chat_bloc.dart
@@ -20,10 +20,21 @@ class ChatBloc {
   static StreamSubscription<NewMessageEvent>? _messageSubscription;
 
   final RxList<Chat> _chats = <Chat>[].obs;
-  RxList<Chat> get chats => _chats;
+
+  RxList<Chat> get chats => _chats
+      .where((e) => SettingsManager().settings.skin.value == Skins.iOS ? !(e.isPinned ?? false) : true)
+      .toList()
+      .obs;
+
+  RxList<Chat> get pinnedChats => _chats
+      .where((e) => SettingsManager().settings.skin.value == Skins.iOS ? e.isPinned ?? false : false)
+      .toList()
+      .obs;
   final RxInt _unreads = 0.obs;
+
   RxInt get unreads => _unreads;
   bool _hasChats = false;
+
   bool get hasChats => _hasChats;
 
   void updateUnreads() {
@@ -41,7 +52,9 @@ class ChatBloc {
   }
 
   int get pageSize {
-    return (SettingsManager().settings.denseChatTiles.value || SettingsManager().settings.skin.value != Skins.iOS) ? 12 : 10;
+    return (SettingsManager().settings.denseChatTiles.value || SettingsManager().settings.skin.value != Skins.iOS)
+        ? 12
+        : 10;
   }
 
   Future<Chat?> getChat(String? guid) async {
@@ -358,12 +371,14 @@ class ChatBloc {
   }
 }
 
-extension Archived on RxList<Chat> {
+extension Helpers on RxList<Chat> {
   /// Helper to return archived chats or all chats depending on the bool passed to it
   /// This helps reduce a vast amount of code in build methods so the widgets can
   /// update without StreamBuilders
   archivedHelper(bool archived) {
-    if (archived) return this.where((e) => e.isArchived ?? false).toList();
-    else return this;
+    if (archived)
+      return this.where((e) => e.isArchived ?? false).toList();
+    else
+      return this.where((e) => !(e.isArchived ?? false)).toList();
   }
 }
--- a/lib/blocs/chat_bloc.dart
+++ b/lib/blocs/chat_bloc.dart
@@ -21,15 +21,7 @@ class ChatBloc {
 
   final RxList<Chat> _chats = <Chat>[].obs;
 
-  RxList<Chat> get chats => _chats
-      .where((e) => SettingsManager().settings.skin.value == Skins.iOS ? !(e.isPinned ?? false) : true)
-      .toList()
-      .obs;
-
-  RxList<Chat> get pinnedChats => _chats
-      .where((e) => SettingsManager().settings.skin.value == Skins.iOS ? e.isPinned ?? false : false)
-      .toList()
-      .obs;
+  RxList<Chat> get chats => _chats;
   final RxInt _unreads = 0.obs;
 
   RxInt get unreads => _unreads;
@@ -375,10 +367,17 @@ extension Helpers on RxList<Chat> {
   /// Helper to return archived chats or all chats depending on the bool passed to it
   /// This helps reduce a vast amount of code in build methods so the widgets can
   /// update without StreamBuilders
-  archivedHelper(bool archived) {
+  RxList<Chat> archivedHelper(bool archived) {
     if (archived)
-      return this.where((e) => e.isArchived ?? false).toList();
+      return this.where((e) => e.isArchived ?? false).toList().obs;
     else
-      return this.where((e) => !(e.isArchived ?? false)).toList();
+      return this.where((e) => !(e.isArchived ?? false)).toList().obs;
+  }
+
+  RxList<Chat> bigPinHelper(bool pinned) {
+    if (pinned)
+      return this.where((e) => SettingsManager().settings.skin.value == Skins.iOS ? (e.isPinned ?? false) : true).toList().obs;
+    else
+      return this.where((e) => SettingsManager().settings.skin.value == Skins.iOS ? !(e.isPinned ?? false) : true).toList().obs;
   }
 }
--- a/lib/layouts/conversation_list/pinned_conversation_tile.dart
+++ b/lib/layouts/conversation_list/pinned_conversation_tile.dart
@@ -243,6 +243,7 @@ class _PinnedConversationTileState extends State<PinnedConversationTile> with Au
         onLongPressStart: (details) {
           _tapPosition = details.globalPosition;
         },
+        onTap: onTapUpBypass,
         onLongPress: () {
           HapticFeedback.mediumImpact();
           showMenu(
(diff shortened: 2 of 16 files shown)


lib/repository/models/settings.dart
```diff
@@ -93,6 +93,12 @@ class Settings {
   final Rx<ThemeMode> theme = ThemeMode.system.obs;
   final Rx<SwipeDirection> fullscreenViewerSwipeDir = SwipeDirection.RIGHT.obs;
 
+  // Pin settings
+  final RxInt pinRowsPortrait = RxInt(3);
+  final RxInt pinColumnsPortrait = RxInt(3);
+  final RxInt pinRowsLandscape = RxInt(1);
+  final RxInt pinColumnsLandscape = RxInt(6);
```
tneotia, line 100: Seems like these settings aren't stored in the DB from what I see in the changes. I guess that isn't intended?

lib/blocs/chat_bloc.dart
```diff
@@ -20,10 +20,21 @@ class ChatBloc {
...
+
+  RxList<Chat> get chats => _chats
+      .where((e) => SettingsManager().settings.skin.value == Skins.iOS ? !(e.isPinned ?? false) : true)
+      .toList()
+      .obs;
+
+  RxList<Chat> get pinnedChats => _chats
+      .where((e) => SettingsManager().settings.skin.value == Skins.iOS ? e.isPinned ?? false : false)
+      .toList()
+      .obs;
```
tneotia, line 32: For this I'd prefer if the chat list returned all the chats so that we can use this getter everywhere else besides the conversation list to get all the chats, regardless of pinned status. If you could maybe make another helper function like I did at the bottom I think that would be good, something like this:

```
pinnedHelper(bool pinned) {
    if (pinned)
      return this.where((e) => e.isPinned ?? false).toList();
    else
      return this;
  }

File: lib/blocs/chat_bloc.dart [hunk 3]
```diff
@@ -358,12 +371,14 @@ class ChatBloc {
...
   /// Helper to return archived chats or all chats depending on the bool passed to it
   /// This helps reduce a vast amount of code in build methods so the widgets can
   /// update without StreamBuilders
   archivedHelper(bool archived) {
-    if (archived) return this.where((e) => e.isArchived ?? false).toList();
-    else return this;
+    if (archived)
+      return this.where((e) => e.isArchived ?? false).toList();
+    else
+      return this.where((e) => !(e.isArchived ?? false)).toList();
```
tneotia, line 382: I think the goal here was to return *all* chats when `archived` was false, but it seems you updated it to return non-archived chats? I think we should do it this way but I'm kind of curious why it was working correctly before lol.

tneotia (changes requested):
Couple of questions but otherwise looks great! :)

lib/blocs/chat_bloc.dart
tneotia: You could use that in conjunction with the archived helper to get pinned & archived chats, or make another helper for that specific case.

lib/repository/models/settings.dart
jjoelj: It's intended for now. They're basically just placeholder variables so I don't hardcode values. When I make them changeable I'll add the configentry. Just don't want any broken settings being stored in the db until I'm sure I'm gonna use them. For example, right now the pinColumns in landscape are actually determined by device size, not the variable here.

So basically, until I properly support different device sizes, these are just dummy variables.

lib/blocs/chat_bloc.dart
jjoelj: The reason it was working before was that we had a check in the main chat list to return a `Container()` if we were in archive mode and the chat wasn't archived or vice versa. This fix eliminated the need for that extra code as well.

This was the code (that I got rid of)
`if (!showArchived && ChatBloc().chats.archivedHelper(showArchived)[index].isArchived!) return Container();
if (showArchived && !ChatBloc().chats.archivedHelper(showArchived)[index].isArchived!) return Container();`

lib/blocs/chat_bloc.dart
jjoelj: The reason I did it this way was that there were no uses of this chats getter outside of conversation_list lol. Will make this change tho

tneotia: > The reason I did it this way was that there were no uses of this chats getter outside of conversation_list lol. Will make this change tho

Yea in the current state it's like that but once we start consolidating all of the blocs / managers we may end up using it elsewhere since we want everything to be reactive and update state as things change. Shouldve been more clear ab that

lib/blocs/chat_bloc.dart
tneotia: Ahhh good looks thanks!

lib/blocs/chat_bloc.dart
jjoelj: Done.

lib/repository/models/chat.dart
```diff
@@ -835,6 +835,11 @@ class Chat {
         [this.guid]);
   }
 
+  Future<Rx<Message>> get latestMessage async {
+    List<Message> latests = await Chat.getMessages(this, limit: 1);
+    return latests.first.obs;
```
tneotia, line 840: This won't do anything meaningful because you don't ever set latest.first to a different value. I think it will work the same way if you just get rid of the reactive state here entirely (no .obs and no Rx).

lib/layouts/conversation_list/pinned_conversation_tile.dart
```diff
@@ -0,0 +1,486 @@
...
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_svg/svg.dart';
+import 'package:get/get.dart';
+
+class PinnedConversationTile extends StatefulWidget {
+  final Chat chat;
+  final bool? onTapGoToChat;
```
tneotia, line 29: Can we make this non nullable? I'd prefer all bools to be strictly true/false

lib/layouts/conversation_list/pinned_conversation_tile.dart
```diff
@@ -0,0 +1,486 @@
...
+class _PinnedConversationTileState extends State<PinnedConversationTile> with AutomaticKeepAliveClientMixin {
+  bool isFetching = false;
+  Brightness? brightness;
+  Color? previousBackgroundColor;
+  bool gotBrightness = false;
+
+  // Typing indicator
+  RxBool showTypingIndicator = false.obs;
+
+  void loadBrightness() {
```
tneotia, line 56: I could be wrong but from what I remember migrating settings this bit of code doesn't actually do anything... We may be able to remove this function entirely.

lib/layouts/settings/chat_list_panel.dart
```diff
@@ -211,15 +217,47 @@ class ChatListPanel extends StatelessWidget {
...
+                      return SizedBox.shrink();
+                  }),
+                  Obx(() {
+                    if (SettingsManager().settings.skin.value == Skins.iOS)
+                      return SettingsSlider(
+                        min: 2,
+                        max: 4,
+                        divisions: 3,
+                        update: (double val) {
+                          SettingsManager().settings.pinRowsPortrait.value = int.parse(val.toStringAsFixed(0));
```
tneotia, line 241: Couldn't you just do `val.toInt()` here?

tneotia (changes requested):
Couple of minor things. It looks and feels awesome btw!

lib/layouts/conversation_list/pinned_conversation_tile.dart
zlshames: I agree

zlshames:
There are also merge conflicts with the latest stuff that @tneotia pushed for showing progress indicators when the chat bloc is loading

lib/repository/models/chat.dart
jjoelj: Ah ok good to know. I was just trying some stuff because the speech bubble was occasionally not showing up on a new message. One of the changes in the latest commit fixed it but ig not this one lol.

lib/layouts/conversation_list/pinned_conversation_tile.dart
jjoelj: Since that function is in literally every widget, would it be better to remove them all at once in a different pr?

lib/layouts/conversation_list/pinned_conversation_tile.dart
jjoelj: What is this variable for? It's not used anywhere. Neither is onTapCallback

zlshames: @jjoelj it's used in Material theme, when selecting multiple chats, I think?

lib/layouts/conversation_list/pinned_conversation_tile.dart
tneotia: Heh probably didn't realize we used it everywhere

tneotia (approved):
LGTM

zlshames:
Branch has conflicts. Make sure to fix those, and then lmk when this is ready to be tested

lib/layouts/conversation_list/pinned_conversation_tile.dart
jjoelj: Nah @tneotia there's `inSelectMode` and `onSelect` for that. These have 0 code occurrences lol

zlshames: > Nah @zlshames there's `inSelectMode` and `onSelect` for that. These have 0 code occurrences lol

Well then i guess it's not used anymore. You can remove it

jjoelj: @zlshames already done

jjoelj:
Ready to merge

PR #1239, opened 2021-08-03
Add loading indicator for getting chats on startup

### LINKED ISSUE #1238: Add progress loader to new chat creator when the ChatBloc is still loading
Should be very easy to do. That's why it's high priority. Plus, good for UX

tneotia:
Fixes #1238

### DIFF
--- a/lib/blocs/chat_bloc.dart
+++ b/lib/blocs/chat_bloc.dart
@@ -25,6 +25,8 @@ class ChatBloc {
   RxInt get unreads => _unreads;
   bool _hasChats = false;
   bool get hasChats => _hasChats;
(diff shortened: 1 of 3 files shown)


zlshames (changes requested):
I started doing this, but looks like you finished before me. Can you create a UI helper for getting a loader? It should take 1 param and 2 optional params. The 1 param is the Build context. The other 2 is height/width, which should default to 20 a piece. In the function, you will check the skin if it's iOS. If so, return the ThemeData wrapped Cupertino activity indicator. And make the radius half the passed in width. For material, you can use height/width normally

tneotia:
@zlshames done

zlshames (approved):

PR #1246, opened 2021-08-03
Compress attachment immediately after downloading

### LINKED ISSUE #1106: Compress images on download instead of on load

tneotia:
Fixes #1106

### DIFF
--- a/lib/helpers/attachment_downloader.dart
+++ b/lib/helpers/attachment_downloader.dart
@@ -99,8 +99,6 @@ class AttachmentDownloader {
       debugPrint("Attachment downloaded in ${stopwatch.elapsedMilliseconds} ms");
 
       try {
-        // Get the dimensions of the attachment
-        await AttachmentHelper.setDimensions(attachment);
         // Compress the attachment
         await AttachmentHelper.compressAttachment(attachment, attachment.getPath());
         await attachment.update();
--- a/lib/helpers/attachment_downloader.dart
+++ b/lib/helpers/attachment_downloader.dart
@@ -101,6 +101,8 @@ class AttachmentDownloader {
       try {
         // Get the dimensions of the attachment
         await AttachmentHelper.setDimensions(attachment);
+        // Compress the attachment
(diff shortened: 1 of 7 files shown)


zlshames (changes requested):
Crap, this is my fault originally. But we need to make compressAttachment only handle image file-types, right? Like exclude video (and gifs?). That way it doesn’t try to compress EVERY attachment. Similar to what we do in `setDimensions`. I’m also wondering if we should combine `setDimensions` and `compressAttachment`

tneotia:
Fixed that

zlshames (changes requested):
Sorry to be nitpicky, but I would like to combine them and get rid of setDimensions. Looks like setDimensions uses ImageSizeGetter. If we can remove that dependency and just use FlutternativeImage, that's better imo. We still need to check if it's a GIF, and if so, use the GIF image dimensions call. And if it's a video, use the getVideo dimensions call. But we should essentially combine them imo

tneotia:
Done

zlshames (changes requested):
Hey, so can you call the ImageSizeGetter in 2 cases? 1, in the catch (as you already have it). And 2, if the height/width returned by FlutterNativeImage are 0 (either of them).

So we still need to do the 2nd case. I just want to make sure that we get the size no matter what (to our best ability)

tneotia:
> Hey, so can you call the ImageSizeGetter in 2 cases? 1, in the catch (as you already have it). And 2, if the height/width returned by FlutterNativeImage are 0 (either of them).
> 
> So we still need to do the 2nd case. I just want to make sure that we get the size no matter what (to our best ability)

Done

PR #1287, opened 2021-08-05
Potentially fix sharing to a chat that is already open

### LINKED ISSUE #1275: Images shared to a chat that is already opened, do not get added to the share preview
_No content_

>  Issue created by BlueBubbles Bot in the `BlueBubbles` commnity on behalf of `@zach` in `#community-contributions`.

tneotia:
Fixes #1275, couldn't test bc my share targets wouldn't update


zlshames:
> Fixes #1275, couldn't test bc my share targets wouldn't update

FYI, I don’t think this is unique to share targets. I think the issue happens when sharing normally as well

zlshames:
Merging this, but leaving the ticket open until we confirm the fix

tneotia:
> > Fixes #1275, couldn't test bc my share targets wouldn't update
> 
> FYI, I don’t think this is unique to share targets. I think the issue happens when sharing normally as well

When we share normally it opens the new chat creator - if you're in a chat it shouldn't affect anything, at least it hasn't for me.

jjoelj:
can confirm doesn't work

tneotia:
Rip

PR #1300, opened 2021-08-06
Prioritize contacts with avatars

jjoelj:


lib/layouts/widgets/contact_avatar_group_widget.dart
```diff
@@ -38,311 +48,313 @@ class _ContactAvatarGroupWidgetState extends State<ContactAvatarGroupWidget> {
...
-                        decoration: BoxDecoration(
-                          borderRadius: BorderRadius.circular(widget.size / 2),
-                          color: context.theme.accentColor.withOpacity(0.6),
+    return Obx(
+      () => Container(
+        width: widget.size,
+        height: widget.size,
+        child: participants.length > 1
+            ? ThemeSwitcher(
+                iOSSkin: Obx(
```
tneotia, line 57: I don't really think the nested obx is necessary here now that you have one surrounding the whole thing. I'd try removing it and make sure everything works as intended otherwise add it back I guess.

tneotia (changes requested):

zlshames (approved):

zlshames (approved):

lib/layouts/widgets/contact_avatar_group_widget.dart
jjoelj: Got it to work without

tneotia (approved):
As long as you verified removing the Obx didn't break anything, LGTM :)

PR #1322, opened 2021-08-07
Fix url regex to not allow whitespace in urls

jjoelj:


zlshames (approved):

zlshames:
I merged this, but we definitely want to allow non alpha numeric characters... Are u not allowing #@$-, etc?

jjoelj:
> I merged this, but we definitely want to allow non alpha numeric characters... Are u not allowing #@$-, etc?

The functionality is the same as the original. It doesn't allow nonalphanumeric in the actual base url. E.g. www.@%&!.com won't be a url. Google.com/#~*~&!* is still a url tho.

PR #1341, opened 2021-08-13
Stop url regex from recognizing emails. Support more urls. Ensure url…

jjoelj:
… starts and ends without whitespace.


zlshames (changes requested):
Does this allow when a message starts or ends with a link? Like we don't want to ignore starts of lines

jjoelj:
> Does this allow when a message starts or ends with a link? Like we don't want to ignore starts of lines

Yep. It checks if the link is preceded either by whitespace or the start of the string, and if it ends before whitespace or the end of the string.

PR #1366, opened 2021-08-16
Increase the default preview quality for everyone

tneotia:


zlshames (changes requested):
I don't like the idea of just adding 25 to it. TBH I would just remove the DB Upgrade item. I don't think it should be needed. It'll be set to 50 for new installs. Everyone else will be "grandfathered" into their current preview quality

tneotia:
@zlshames done

zlshames (approved):

PR #1384, opened 2021-08-18
Get pinned chats first

tneotia:


zlshames:
Did you notice any performance differences when doing this?

tneotia:
Nothing noticeable for me because I don't have a ton of chats but I wouldn't be surprised if it is slightly slower for those who have a ton of chats. Don't know how else we could accomplish loading pinned chats first though unless we store chats themselves differently or something (like maybe order the table by pinned chats when making a chat pinned)?

PR #1392, opened 2021-08-23
Improve URL preview to recognize messages with links and text & rework splitting of links when sending a new message

tneotia:


lib/action_handler.dart
```diff
@@ -45,31 +45,17 @@ class ActionHandler {
...
     // Check for URLs
-    RegExpMatch? linkMatch;
-    String? linkMsg;
-    List<RegExpMatch> matches = parseLinks(text);
-
-    // Get the first match (if it exists)
-    if (matches.length > 0) {
-      linkMatch = matches.first;
-      linkMsg = text.substring(linkMatch.start, linkMatch.end).trim();
+    int index = text.indexOf("https://");
```
zlshames, line 48: What if the link is a `www.` link

zlshames (changes requested):
Can you elaborate how it's changed? Like what it changed from -> to

lib/action_handler.dart
tneotia: AM doesn't split it so I didn't add it lol but here's some examples:

`testhttps://google.com` -> `test` and `https://google.com` (same for http links)

`testhttps://google.com test` -> `test` and `https://google.com test` (same for http links)

`testwww.google.com` stays the same (along with any other non http / https links

zlshames: I'll be honest, i don’t want it to be able to pick up `testhttps://google.com test`. To me, that should be an invalid link and not be "preview-able". IMO, links should either be considered links if the following is present:

- Line starts with https://, http://, or, www
- a space, followed by https://, http://, or, www.

I don’t think we should support when it's just a substring

tneotia: Alright so I'll get rid of the link splitting entirely as well as the detection algo for improperly formatted links and we should be good

tneotia: @zlshames this should be good now

PR #1399, opened 2021-08-23
Remove image data when app is no longer in foreground to reduce the memory footprint

tneotia:


zlshames:
Does this have any impact on load times of images when it's brought back from the foreground?

tneotia:
Not that I've noticed

PR #1436, opened 2021-08-31
Fix incoming message color getting reset, photo viewer not working, add better logging, incremental sync pagination

### LINKED ISSUE #723: Incoming messages aren't colored by Custom Avatar Color
and it doesnt recolor previous messages
you have to close the convo and reopen it for previous messages from the contact to get recolored

### LINKED ISSUE #1409: Better logging mechanisms
I want to create a log handler. One where we can control how the logs are written, and if the logs are written. Then we can get users to send us that log for debugging

### LINKED ISSUE #1419: Make incremental sync fetch messages in batches
Right now it fetches the last 1k all at once and then ingests them in batches. We should fetch the messages in batches so the initial requests take a shorter time to bring in the newest messages

### LINKED ISSUE #1435: Double-tap to zoom in full screen attachment viewer interferes with left/right swipe actions
Overall, our image preview gestures kinda suck and need to be looked at/fixed. Zooming with 2 fingers is difficult, and double-tap to zoom works, but then moving around while zoomed does not

tneotia:
Fix #723, fix #1435, fix #1409, fix #1419

### DIFF
--- a/lib/action_handler.dart
+++ b/lib/action_handler.dart
@@ -19,6 +19,7 @@ import 'package:bluebubbles/managers/settings_manager.dart';
 import 'package:bluebubbles/repository/database.dart';
 import 'package:bluebubbles/repository/models/attachment.dart';
 import 'package:bluebubbles/repository/models/chat.dart';
+import 'package:bluebubbles/repository/models/handle.dart';
 import 'package:bluebubbles/repository/models/message.dart';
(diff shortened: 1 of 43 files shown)


zlshames (changes requested):
Can we make the batches for incremental sync 100 instead of 10

PR #1555, opened 2021-09-17
Fix contacts sync on desktop / web

tneotia:


tneotia:
Implemented in webapp rc1 PR

PR #1676, opened 2021-11-07
Generate notification messages from associated message

elliotnash:
This pr fixes reaction notifications being in the wrong language when sent from a device in another language.

### DIFF
--- a/lib/helpers/message_helper.dart
+++ b/lib/helpers/message_helper.dart
@@ -1,5 +1,7 @@
 import 'dart:async';
 import 'package:bluebubbles/blocs/chat_bloc.dart';
+import 'package:bluebubbles/helpers/reaction.dart';
+import 'package:bluebubbles/repository/database.dart';
 import 'package:flutter/foundation.dart';
 import 'package:google_ml_kit/google_ml_kit.dart';
 import 'package:universal_html/html.dart' as html;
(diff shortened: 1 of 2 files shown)


lib/helpers/message_helper.dart
```diff
@@ -318,80 +290,76 @@ class MessageHelper {
...
+      if (associatedMessage != null) {
+        // grab the verb we'll use from the reactionToVerb map
+        String? verb = ReactionTypes.reactionToVerb[message.associatedMessageType];
+        // we need to check balloonBundleId first because for some reason
+        // game pigeon messages have the text "�"
+        if (associatedMessage.isInteractive()) {
+          return "$sender $verb ${getInteractiveText(associatedMessage)}";
+        // now we check if theres a subject or text and construct out message based off that
+        } else if (isNullOrEmpty(associatedMessage.text, trimString: true) == false
+            || isNullOrEmpty(associatedMessage.text, trimString: true) == false) {
```
tneotia, line 309: This condition can be shortened to `!isNullOrEmpty(associatedMessage.text, trimString: true) || !isNullOrEmpty(associatedMessage.text, trimString: true)`

tneotia (changes requested):
Minor change but otherwise great stuff! :)

lib/helpers/message_helper.dart
elliotnash: isNullOrEmpty returns an optional bool (not entirely sure why) so I'd have to unwrap the bool first, I'm not sure if its okay to assert isNullOrEmpty as not null

tneotia: Ah I forgot about that, yes you can assert it as not null. It should never return null, but for some reason one of the packages we use has its own `nullOrEmpty` functions return null for whatever reason.

tneotia (approved):
LGTM :)

zlshames:
Good work, thanks for the contribution!

PR #1681, opened 2021-11-11
Remove Sammy theme for now (1.7.0 Final Changes)

tneotia:


lib/layouts/settings/theme_panel.dart
```diff
@@ -166,7 +166,7 @@ class ThemePanel extends GetView<ThemePanelController> {
                         saveSettings();
                         controller.update();
                       },
-                      options: Skins.values.toList(),
+                      options: Skins.values.where((item) => item != Skins.Samsung).toList(),
```
zlshames, line 169: didn’t you say this could cause an error for people who have it enabled already or no?

tneotia: yes, because it'll try to find the selected item as `Skins.Samsung` but that doesn't appear in the options list.

zlshames: Should we fix that then? or should we add a case where if it's not in the list, default it to something

zlshames:
Are you gonna add the reaction notification text fixes into this as well?

tneotia:
Yep gimme just a few secs and I'll push

lib/layouts/settings/theme_panel.dart
tneotia: Ehh idk since its a change from alphas, not stable. But I might be able to cook up something later today around 5 ish

tneotia:
@zlshames done

lib/layouts/settings/theme_panel.dart
tneotia: Cooked something up rq so we should be good here

PR #1819, opened 2021-12-25
[WIP] 1.8.0 Stable

tneotia:


lib/helpers/logger.dart
```diff
@@ -76,7 +77,7 @@ class BaseLogger extends GetxService {
         onPressed: () {
           Share.file("BlueBubbles Logs", filePath);
         },
-        child: Text("SHARE", style: TextStyle(color: Theme.of(Get.context!).primaryColor)),
+        child: kIsDesktop || kIsWeb ? Container() : Text("SHARE", style: TextStyle(color: Theme.of(Get.context!).primaryColor)),
```
tneotia, line 80: @jjoelj instead of doing this couldn't we have just made the `button` param null when `kIsDesktop || kIsWeb`?

PR #1888, opened 2022-02-21
Update pluginhandle to return 0 if no isolate value set

mrsylerpowers:
Currently, the code returns -1 if no isolate set but adjusting it the shared prefs to return 0 if not found as the second parameter of `getLong` is the default parameter if not found

### DIFF
--- a/android/app/src/main/java/com/bluebubbles/messaging/services/FlutterFirebaseMessagingBackgroundExecutor.java
+++ b/android/app/src/main/java/com/bluebubbles/messaging/services/FlutterFirebaseMessagingBackgroundExecutor.java
@@ -188,7 +188,7 @@ public class FlutterFirebaseMessagingBackgroundExecutor implements MethodCallHan
             if (callbackHandle != 0) {
                 Log.i(TAG, "Found background isolate callback handle");
                 startBackgroundIsolate(callbackHandle, null);
-            }else {
+            } else {
                 Log.i(TAG, "No callback handler available for background isolate");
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/bluebubbles/messaging/services/FlutterFirebaseMessagingBackgroundExecutor.java
```diff
@@ -184,9 +184,12 @@ public void onMethodCall(MethodCall call, @NonNull Result result) {
      */
     public void startBackgroundIsolate() {
         if (isNotRunning()) {
-            Long callbackHandle = getPluginCallbackHandle();
-            if (callbackHandle != null && callbackHandle != 0) {
+            long callbackHandle = getPluginCallbackHandle();
+            if (callbackHandle != 0) {
+                Log.i(TAG, "Found background isolate callback handle");
                 startBackgroundIsolate(callbackHandle, null);
+            }else {
```
zlshames, line 191: Crazy nitpicky, but can u put a space after this braces

zlshames (changes requested):

android/app/src/main/java/com/bluebubbles/messaging/services/FlutterFirebaseMessagingBackgroundExecutor.java
mrsylerpowers: Done

PR #1942, opened 2022-03-18
CICD: Create Auto-Build Workflows using GitHub Actions

### LINKED ISSUE #1936: Automated builds for desktop
Setup workflows through GitHub actions which automatically build the desktop versions and upload them to a release whenever a tag (beginning with the letter v, as all the current tags seem to be) is created. 

I'm already working on this, just figured I would create a ticket for it so two people aren't working on it.

Morpheus636:
Closes #1936

This is a draft PR for #1936, creating auto-build workflows for the desktop versions of bluebubbles-app. Appologies for the long commit history (at least, long for the size of the actual changes), but unfortunately the only way to test GitHub actions is in production (although I've figured out a better way to do this so the windows workflow should not have as many commits). 

At this point, the PR includes an auto-build workflow for the Linux build which runs in GitHub actions  whenever a new tag beginning with the letter `v` is created. The workflow builds the linux version of the app, zipping the `bundle` directory into `bluebubbles-linux-x86.zip` and uploading it as an asset within the release associated with the tag.

I plan to begin work on the Windows workflow now, but wanted to create the PR so my work can be properly tracked.


war59312:
Abandoned?

jjoelj:
For anyone who wants to continue working on this, here's the gist of what needs to be added/changed:
- [ ] The `onContentCommit` stuff no longer needs to be commented out
- [ ] Linux can use the script `/linux/build.sh` so that the version number gets placed into the binary correctly
- [ ] It would also be helpful if we could set the flutter version from an env variable, since releases are not always made using the latest stable version of Flutter (especially when major versions of Flutter are released a few days before we make a release).
- [ ] The GIPHY_API_KEY needs to be placed in the .env file from a repository secret, which we can add to the repo once whoever takes this over adds it to the workflow.
  - [ ] There's also the AppID in the inno file that might need to be a secret, but I'm not sure about that

I'll add anything else I think of, but this should be the bulk of the work.

jjoelj:
#3046 will add this

PR #1961, opened 2022-04-01
Chat Peek Feature

tneotia:


zlshames:
Is there any way to make it do it utilizes most the screen real-estate so it can at least show the mark as read option all the time (if space available)? I remember using this and i always got the "more" menu entry _except_ for the top row of pinned chats, which is kinda annoying. Especially since we are removing/overriding the tap n hold functionality

tneotia:
@zlshames done

PR #2084, opened 2022-06-01
Check for server updates on app cold start

### LINKED ISSUE #1987: Auto check for server updates every so often & notify user
Do this when the app starts fresh. On a delay so it doesn't hold up the UI and cause lag. Only do it at max 1 time every 6 hours. Add a toggle to turn off reminders for it if users want to ignore them

tneotia:
Closes #1987

### DIFF
--- a/lib/layouts/settings/server_management_panel.dart
+++ b/lib/layouts/settings/server_management_panel.dart
@@ -743,7 +743,6 @@ class ServerManagementPanel extends GetView<ServerManagementPanelController> {
                       materialIcon: Icons.dvr,
                     ),
                     onTap: () async {
-
                       final response = await api.checkUpdate();
                       if (response.statusCode == 200) {
                         bool available = response.data['data']['available'] ?? false;
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -5,6 +5,7 @@ import 'dart:ui';
 
 import 'package:adaptive_theme/adaptive_theme.dart';
 import 'package:bitsdojo_window/bitsdojo_window.dart';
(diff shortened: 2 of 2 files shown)


zlshames:
Looks good, but please make this on a 15 second delay as to not interfere with startup code such as initiating the DB store or startup time.

zlshames (changes requested):
Add delay

tneotia:
This doesn't actually interfere with startup code, the API call isn't `await`ed so other code is free to run after it.

tneotia:
I dunno if its really going to cause lag tbh, its just a network call and doesn't update the UI at all. There isn't any sort of heavy "processing" to it. I can shift the API call over to an isolate if you want, though (but the actual rendering of the dialog will need to be done on the main thread)

zlshames:
> I dunno if its really going to cause lag tbh, its just a network call and doesn't update the UI at all. There isn't any sort of heavy "processing" to it. I can shift the API call over to an isolate if you want, though (but the actual rendering of the dialog will need to be done on the main thread)

How is this shown when you do the following:
1. Tap through a notification
2. App opens and update dialog shows
3. Navigator transitions to the chat opened via the notification

Will there be any issues with navigation/the back button?

zlshames:
> No I dont think so, it just depends which "route" is pushed on the stack first. If it takes a while for the app to open the message page, itll probably open the dialog first, then the message page on top of it. When the user presses back, itll take them back to the convo screen but the dialog will be popped open. Then they hit "OK" to close it.

ok that's fine

PR #2108, opened 2022-06-07
fix icloud account display in server management

jjoelj:
this pr also includes desktop 1.9.5 because I branched off of it. The last commit is the relevant fix


jjoelj:
Contained in 1.9.5

PR #2112, opened 2022-06-09
[1.10] Completely revamp theming system

tneotia:
- Better support Material 3 standards
- Pave the way for adding more theming options easily
- Vastly simplify backend theming code to reduce bugs and make the system easier to work with

NOTE: This change will completely delete themes from previous app versions when upgrading. A migration has not been written yet, and will need to be discussed (to see if it is even possible).
NOTE 2: This commit makes the app look absolutely *terrible*, this will be fixed slowly during UI rewrites


zlshames:
Can you change this to merge with the `feature/bordeaux/theme-rewrite` branch

Issue #2119, opened 2022-06-10
Reduce padding between contact name and message bubble


zlshames:
This is based on how it looks on iOS. The name is closer to the bubble. Make sure this doesn't break other formatting.

tneotia:
Done

PR #2192, opened 2022-07-14
Reduced padding between contact name and message bubble

jakeump:
Added horizontalTitleGap for each skin
#2119


jakeump:
@tneotia I can fix that as well. The horizontal tile spacing was indeed too large compared to iMessage, so regardless it fixes something

tneotia-sims:
OK thanks, I would just check and make sure that compared to google messages and samsung messages the spacing isn't off. We don't want to use the iMessage spacing difference for the other 2 skins if they are different

jakeump:
Closed because I suck at git and don't know how to revert the files. Going to create another branch and do new PR

Issue #524, opened 2020-12-05
Custom incoming/outgoing message sounds


lanes100:
Request: 
Incoming and outgoing sounds having custom sounds in the way of a user selected sound file. Some preselected sounds would be nice too.

zlshames:
This is just a matter of doing 3 things, but overall fairly easy. Marking this as `medium` because it touches on a few parts of the app:

- Adding an audio resource for the send sound
- Adding code here to play the sound: https://github.com/BlueBubblesApp/BlueBubbles-Android-App/blob/v1.0.0/lib/layouts/conversation_view/conversation_view.dart
- Adding a toggle to the settings (this should be off by default)

tneotia:
done

PR #2194, opened 2022-07-16
Play send sound setting for iOS

jakeump:
* Option to play send sound if theme is iOS
* Default iMessage sound plays on send if enabled
* #524


tneotia:
See my comments in discord #community-contributions

zlshames:
As well as mine. Thank you for the PR and work, I just think we can't provide the same assets that Apple uses for legal reasons. An alternative could be that we allow users to choose a custom mp3 to use. As well as have a few presets that are royalty free

jakeump (commented):
updated

tneotia:
Since this includes UI changes in settings, can you rebase off of tanay/bordeaux/widget-rewrites and target feature/widget-rewrites in your PR? Just to avoid a lot of merge conflicts. Sorry for the change!

Issue #2458, opened 2023-04-21
Desktop - Special characters in attachment name cause download to fail


mkitchingh:
Desktop app: 1.12.2
Server Version: 1.7.2
Attachment name was cm-chat-media-video-1:2e618657-1cd2-4dad-9201-7af81daf23e4:2561:0:0.MOV
In this case, the colon caused it to fail. 
Reproduced by manually trying to create the same file. 
```
C:\Users\Ma****in\AppData\Roaming\BlueBubbles\bluebubbles\attachments\012C05B5-7087-467E-A2EA-ED257DBC6F53>echo "hello world"> cm-chat-media-video-1:2e618657-1cd2-4dad-9201-7af81daf23e4:2561:0:0.MOV
The filename, directory name, or volume label syntax is incorrect
```.
The file was created by changing the colons to dashes. In this case it was a colon, but any character that is illegal in a Windows file name should cause the same issue.

PR #2474, opened 2023-05-01
Fix issue #2458

MatthewStadter:


lib/models/io/attachment.dart
```diff
@@ -223,7 +223,7 @@ class Attachment {
 
   String get directory => "$baseDirectory/$guid";
 
-  String get path => "$directory/$transferName";
+  String get path => Platform.isWindows ? "$directory/$transferName".replaceAll(RegExp(r'[<>:"/\|?*]'), "_") : "$directory/$transferName";
```
jjoelj, line 226: - This changes the file structure, which I assume you didn't intend to do.
- Make sure the original path separators are maintained. (i.e. you should only run replace on the `transferName`, or we'll end up with a flat file structure.)
- We should also filter out forward slashes in filenames on Linux for completeness's sake.

jjoelj (changes requested):
Thanks for contributing!

lib/models/io/attachment.dart
MatthewStadter: Hello, I fixed the issues. Thanks

tneotia:
@jjoelj we can do that ourselves :)

tneotia (approved):
LGTM, thanks for your contribution!

jjoelj (approved):

PR #2554, opened 2023-09-17
Abstract backend

TaeHagen:
Not sure if you would be interested in merging this, but this abstraction facilitates integration of other backends such as `rustpush`.

It mostly moves direct HTTP calls out of the UI and into a backend class. It *should* just be a refactor, but note I don't actually have a BB setup so I haven't tested anything with the real BB server, so you would probably want to do that at some point. Though I have tested it with my other backend (rustpush).


tneotia:
Oops... didnt know that would close this. If possible can you reopen the PR and base it on development, now that the hotfix-5 branch is merged in with it?

TaeHagen:
I'm currently working on reversing the algorithms required to authenticate with iMessage. When I return to the fork I'll rebase it and open a new PR.

Thank you!

PR #2650, opened 2024-01-20
Display vetted aliases in server management panel

SeanReg:
Change uses the account info API to query the vetted aliases to display them in the server management panel. Future PR will handle implement aliases-removed event.

### DIFF
--- a/lib/app/layouts/settings/pages/server/server_management_panel.dart
+++ b/lib/app/layouts/settings/pages/server/server_management_panel.dart
@@ -28,6 +28,7 @@ import 'package:qr_flutter/qr_flutter.dart';
 import 'package:universal_html/html.dart' as html;
 import 'package:universal_io/io.dart';
 import 'package:version/version.dart';
+import 'package:bluebubbles/utils/logger.dart';
 
 class ServerManagementPanelController extends StatefulController {
   final RxnInt latency = RxnInt();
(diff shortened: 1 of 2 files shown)


tneotia:
Thanks for this! Although not explicitly outlined in a FR, I had an idea of how I wanted this done - essentially the "profile" section would now open a new screen upon tapping that shows all this information (vetted alias, currently set alias, ability to change PFP and name, set a different default alias, and check iMessage connectivity). If possible, can you change this PR to start that with just displaying vetted aliases?

SeanReg:
Sure, PR has been updated. Moved account info into alert dialog with button below the "show stats" button

tneotia (changes requested):
Minor thing:

lib/app/layouts/settings/pages/server/server_management_panel.dart
```diff
@@ -77,20 +80,42 @@ class ServerManagementPanelController extends StatefulController {
...
-        opacity.value = 1.0;
-      });
+          })
+          .catchError((e) { Logger.debug("Failed to get account info $e"); })
+        );
+      }
+
+      Future.wait(subsequentRequests)
+        .then((value) {})
+        .catchError((_) { })
```
tneotia, line 117: Are these necessary? If we catchError here, not sure that it will get propagated to line 120.

tneotia (changes requested):

lib/app/layouts/settings/pages/server/server_management_panel.dart
SeanReg: Will remove

tneotia (approved):
LGTM

PR #2652, opened 2024-01-20
Fixes #2651 - replies in notifications not sending

### LINKED ISSUE #2651: [Android] Replies from notifications sporadically don't send
If the flutter engine is closed, replies from notifications won't send (even though they appear to be successful in the UI)

gsxdsm:
Update the worker logic to stop the reply worker from hanging when sending a reply to a notification.


tneotia:
Merging in to get you some credit, but I'll be changing this in the Kotlin impl.

PR #2657, opened 2024-01-24
Updates for aliases removed notification

SeanReg:
- Use new event type string
- Support event over socket
- Avoid duplicate notification (if same event comes from socket + FCM)

### DIFF
--- a/lib/services/backend/notifications/notifications_service.dart
+++ b/lib/services/backend/notifications/notifications_service.dart
@@ -683,7 +683,7 @@ class NotificationsService extends GetxService {
         }
 
         await flnp.show(
-          -3,
+          notifId,
           title,
           text,
           NotificationDetails(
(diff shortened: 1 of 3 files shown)


lib/services/backend/notifications/notifications_service.dart
```diff
@@ -651,21 +652,36 @@ class NotificationsService extends GetxService {
...
     } else {
+        final notifs = await flnp.getActiveNotifications();
+
+        //Already have this notification
+        if (notifs.firstWhereOrNull((n) => n.id == notifId && n.body == text) != null) {
+          return;
+        }
+
         await flnp.show(
           -3,
```
tneotia, line 686: Can you use `notifId` here as well

tneotia (changes requested):

tneotia (commented):
LGTM

PR #2666, opened 2024-02-04
Feature: Check server health in the background

SeanReg:
Optional feature (disabled by default in server management panel). Periodically check server health in the background with a periodic worker (run every 30 minutes when there's network). Shows the server connection error notification when health check fails, clears it when it comes back.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -151,6 +151,7 @@ dependencies {
     implementation 'androidx.browser:browser:1.7.0'
     implementation 'androidx.activity:activity-ktx:1.8.2'
     implementation "androidx.work:work-runtime:2.9.0"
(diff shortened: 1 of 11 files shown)


tneotia (changes requested):
Overall, a good start but some things that need to be fleshed out more. 

I am internally (slowly) working on a persistent socket connection service, but within Kotlin, maybe you can wait for that to be finished so you get some inspiration? It'll be hardened as a persistent service, so start on boot, restart itself if it gets killed, and create a persistent notification that its active, all which we would want this service to do as well. Not sure when I'll be able to finish that, but hopefully in the next few weeks.

war59312:
Abandoned?

zlshames:
Closing this. If you think that you want to re-implement it against `master` (post-rewrite), feel free.

PR #2713, opened 2024-04-07
Fix race condition when delivering background messages

TaeHagen:
https://discord.com/channels/640229510662455326/781023536168304641/1226580773348577401

The first task that spins up an isolate can send the task before the isolate is listening, leading to the task never starting or completing and the isolate running indefinitely. Subsequent messages are handled by the indefinitely running isolate just fine.


android/app/build.gradle
```diff
@@ -151,6 +151,7 @@ dependencies {
     implementation 'androidx.browser:browser:1.7.0'
     implementation 'androidx.activity:activity-ktx:1.8.2'
     implementation "androidx.work:work-runtime:2.9.0"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-guava:1.6.0"
```
tneotia, line 154: Do the coroutines imports in 164-166 not offer the functionality you need? If not then can you move this impl to that coroutines section

tneotia (approved):
Just one question there but otherwise LGTM

PR #2781, opened 2024-08-10
Improved performance of scrolling

funkemunky:
The following indexes should improve the rendering performance on some pages with complex child rebuilds.

Copperboy100 in the discord found the "findChildIndexCallback" option that instructs flutter to find a previously indexed build state, instead of recalculating all the children unnecessarily.


zlshames:
I've actually already implemented this + an improvement on it to prevent type errors in my dev branch: https://github.com/BlueBubblesApp/bluebubbles-app/blob/zach/feat/true-foreground-service/lib/helpers/ui/ui_helpers.dart#L666-L683

this is the helper i wrote that gets implemented on all of the ListView widgets in the app. In addition, this PR has some weird changes like references to dependencies in a user's downloads folder. That said, since I've already implemented this, it's not needed. Apologies!

PR #2798, opened 2024-09-07
Implement UnifiedPush

### LINKED ISSUE #2274: UnifiedPush support
As this is supposed to be a (F)OSS/self-hosted solution, UnifiedPush support as an alternative to Firebase Cloud Messaging would be greatly appreciated

Garland-g:
This PR implements UnifiedPush support and adds a settings page to enable and disable it.

After enabling, a target URL will be displayed in the settings. One more step is required: Open the Bluebubbles Server settings, and add a Webhook that sends New Messages events to the target URL. Push notifications will work, even after the app is killed.

On the latest release (v1.13.2), the socket service accounted for ~33% of the battery usage on mostly-idle Pixel 8 Pro. When using this branch, battery usage dropped to < 1% while receiving push notifications.

Closes #2274


cameronaaron:
In the latest version they are adding in forground service which might make this moot

zlshames:
Sounds to me

> In the latest version they are adding in forground service which might make this moot

Sounds to me like this is a firebase replacement more than a background service replacement. Maybe I'm wrong? But it seems like yeah, the current foreground service implementation drains battery fast, but I'm not sure how much better the true background service changes I made would improve on that. I think it does significantly, but maybe unified push is still more efficient?

I still need to look at the code changes, so..

Garland-g:
> Sounds to me like this is a firebase replacement

Precisely.

No bluebubbles code needs to run in the background at all when either FCM or UnifiedPush is being used. In both cases a separate service listens for incoming push notifications and distributes notifications to the correct applications.

android/app/build.gradle
zlshames, line 1: Any reason why you are referencing a specific commit here? Maybe I'm mistaking the SHA prefix for something else... But if rather not reference a specific commit and rather, the latest stable branch

zlshames (commented):
At a quick glance, the code looks all good. I'll need to do a full review of it when I have some more time. Will definitely try to get this merged for the next next release

android/app/build.gradle
Garland-g: There wasn't a recent release tag listed when I imported the library.

While I was testing this they added a new release that I didn't see. I've made the change to import that release.

zlshames (approved):

zlshames (approved):

zlshames:
@Garland-g any idea why im getting these build errors after merging?

```
* What went wrong:
Could not determine the dependencies of task ':unifiedpush_android:compileReleaseKotlin'.
> Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=8, vendor=any, implementation=vendor-specific} for MAC_OS on x86_64.
   > No locally installed toolchains match and toolchain download repositories have not been configured.

* Try:
> Learn more about toolchain auto-detection at https://docs.gradle.org/8.9/userguide/toolchains.html#sec:auto_detection.
> Learn more about toolchain repositories at https://docs.gradle.org/8.9/userguide/toolchains.html#sub:download_repositories.
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 11s
Running Gradle task 'assembleProdRelease'...                       11.4s
Gradle task assembleProdRelease failed with exit code 1
```

`Could not determine the dependencies of task ':unifiedpush_android:compileReleaseKotlin'.`

zlshames:
I _think_ i was able to resolve it by adding a toolchain resolver into my `settings.gradle` file's `plugins` section:

```
id 'org.gradle.toolchains.foojay-resolver-convention' version '0.8.0'
```

Not sure if you had some sort of other toolchain installed or other...

Garland-g:
It looks like it's asking for Java 8. I have that installed locally.

I came across this thread at some point earlier, and I think that was what fixed it for me.
https://github.com/nextcloud/neon/issues/1810

zlshames:
Well then im confused because I was able to build the app without Java 8 (Java 17), so long as I added that foojay resolver.

Side thing... @Garland-g How tf do I use UnifiedPush lmao. Can you point me to a guide for setting it up and what not?

Garland-g:
I'm not entirely sure about the resolver, but I presume it downloads the Java 8 SDK for you.

The documentation is here.
https://unifiedpush.org/

There are different providers (distributors) for it:
I'm using a self-hosted ntfy server for this, but it might be simplest to use UP-FCM for testing this. That one will use FCM, so it should be usable with the standard setup.

Once you have a distributor installed, you enable Unified Push in the Bluebubbles settings, and it will give you an endpoint for a webhook.

Right now, you have to manually add a webhook in the Bluebubbles server. I'd like to streamline that so that the android app can create the webhook on the server, but I didn't see an endpoint to allow creating a webhook via an API.

PR #2839, opened 2024-11-07
[WIP] Implement fallback server functionality in SocketService

cameronaaron:
Introduce a secondary server URL for socket connections, allowing for automatic switching between primary and fallback servers in case of connection issues. Update relevant configurations and settings to support this feature.

### DIFF
--- a/android/.project
+++ b/android/.project
@@ -16,12 +16,12 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1599102953393</id>
+			<id>1731019137326</id>
 			<name></name>
 			<type>30</type>
 			<matcher>
 				<id>org.eclipse.core.resources.regexFilterMatcher</id>
-				<arguments>node_modules|.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
+				<arguments>node_modules|\.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
 			</matcher>
 		</filter>
 	</filteredResources>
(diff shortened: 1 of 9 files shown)


android/.project
zlshames, line 1: You added these to the gitignore, but still included them in your commit

android/.settings/org.eclipse.buildship.core.prefs
zlshames, line 1: You added these to the gitignore, but still included them in your commit

lib/services/network/socket_service.dart
```diff
@@ -155,39 +164,35 @@ class SocketService extends GetxService {
...
         Logger.info("Disconnected from socket...");
         state.value = SocketState.disconnected;
-        return;
+        break;
       case SocketState.connecting:
         Logger.info("Connecting to socket...");
         state.value = SocketState.connecting;
-        return;
+        break;
       case SocketState.error:
```
zlshames, line 176: I'm not sure i liked that we changed the functionality of fetching from firebase. Especially when not everyone will have a secondary address entered.

Can we maintain the old functionality based on whether or not there is a secondary server address set? If not set, fetch new url from firebase. I also think that if both server URLs are tried before it works, it should also fetch the new url from firebase. Essentially, keep a counter of errors/failures. It's incremented on error, and reset on connection. If errors == serverAddresses.length, then also try fetching new url from firebase

zlshames (changes requested):

zlshames:
Also, @cameronaaron i assume that you didn't implement any settings UI to configure this secondary URL? Just the code to handle it?

cameronaaron:
Exactly this is all a work in progress! I just started it and made the base pull request so that I will actually finish it lol

zlshames:
ok, in that case, i'll hold off on merging it for this next release

lib/services/network/socket_service.dart
cameronaaron: > I'm not sure i liked that we changed the functionality of fetching from firebase. Especially when not everyone will have a secondary address entered.
> 
> Can we maintain the old functionality based on whether or not there is a secondary server address set? If not set, fetch new url from firebase. I also think that if both server URLs are tried before it works, it should also fetch the new url from firebase. Essentially, keep a counter of errors/failures. It's incremented on error, and reset on connection. If errors == serverAddresses.length, then also try fetching new url from firebase



Ah yes I agree that is a better method will do that

PR #2842, opened 2024-11-07
Remeber reply state

cameronaaron:


cshariq (approved):
Compiled with this commit, works fine.

cameronaaron:
> Compiled with this commit, works fine.

This pull is a work in progress. Are you having any bugs or noticing any benefits ??

lib/services/ui/chat/conversation_view_controller.dart
```diff
@@ -232,4 +233,27 @@ class ConversationViewController extends StatefulController with GetSingleTicker
     cm.setAllInactiveSync();
     Get.delete<ConversationViewController>(tag: tag);
   }
-}
\ No newline at end of file
+
+  Future<void> saveReplyToMessageState() async {
+    final prefs = await SharedPreferences.getInstance();
```
zlshames: Pretty sure we already have an instance of shared preferences in the settings service. Use `ss.prefs` to access it.

lib/services/ui/chat/conversation_view_controller.dart
```diff
@@ -232,4 +233,27 @@ class ConversationViewController extends StatefulController with GetSingleTicker
...
+      await prefs.setString('replyToMessage_${chat.guid}', replyToMessage!.item1.guid!);
+      await prefs.setInt('replyToMessagePart_${chat.guid}', replyToMessage!.item2);
+    } else {
+      await prefs.remove('replyToMessage_${chat.guid}');
+      await prefs.remove('replyToMessagePart_${chat.guid}');
+    }
+  }
+
+  Future<void> loadReplyToMessageState() async {
+    final prefs = await SharedPreferences.getInstance();
```
zlshames: Same comment about the shared prefs instance here

zlshames (changes requested):

lib/services/ui/chat/conversation_view_controller.dart
cameronaaron: Dope Will do!

zlshames (approved):

PR #2928, opened 2025-05-16
opt-in to MTE async mode

cameronaaron:
currently only benefits GrapheneOS users on Pixel 8 and 9 who don't turn on MTE for all apps. hopefully will be extended to all Pixel users and then more devices

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -56,6 +56,7 @@
     <application
         tools:replace="android:label"
         android:allowBackup="false"
+        android:memtagMode="async"
         android:name="${applicationName}"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name_en"


zlshames:
I'm not familiar with GrapheneOS or Molly. I understand what turning this flag on to async does, but how is it useful in other ways than security (at the cost of performance) for users? Or is that the main reason to do it?

cameronaaron:
It is mostly a security thing. Performance I would not estimate would be much. In stock Android 16 which should be coming out this month there is a feature that if this flag is in the app and users have opted into advanced protection and are using a handset that supports it they can use memory tagging. So it is sort of an Android 16 security enhancement thing

cameronaaron:
The performance impact here is minimal—it’s unlikely any user would notice it.
While it’s definitely a win for security, it also helps with debugging by making memory corruption issues easier to catch.
Google will likely expand MTE beyond just Advanced Protection, and could eventually require all apps to opt in. Preparing for that now by ensuring the app is free of memory corruption benefits security, debugging, and future compatibility.
Also worth noting: async mode has much lower performance overhead, while sync mode in userspace offers little practical benefit.

cameronaaron:
I made a similar patch for mulvad vpn that was merged in

zlshames:
> I made a similar patch for mulvad vpn that was merged in

I believe you :) I just need to have a moment to finalize my server changes & release them. Then go work on app updates.

cameronaaron:
Ok! Just let me know if you have questions :)

zlshames (approved):

PR #2932, opened 2025-05-26
Fix: Support headless notification channel creation for UnifiedPush

applesoff:
### Summary

This PR resolves an issue where notification channels could not be created in headless contexts (e.g., when triggered by UnifiedPush). Previously, calls to `createNotificationChannel` from Dart would silently fail if the Flutter engine wasn't initialized.

### Changes

- Introduced native handling of `createNotificationChannel` via a dedicated `notification_plugin` MethodChannel.
- Registered this MethodChannel in `MainActivity.kt` so it is always available when the Flutter engine is running.
- Added `NotificationUtils.kt` to safely create notification channels on Android 8.0+.
- Ensured compatibility with UnifiedPush by making it safe to call from background/isolated Dart contexts.

### Testing

I do not have the capability to test this. these changes were generated with AI

### Motivation

UnifiedPush can send messages when the Flutter app is not running in the foreground. For those scenarios, Android still requires a notification channel to be created before sending foreground-style notifications. This fix enables full compliance with Android notification requirements even in background contexts.

### DIFF
--- a/android/app/src/main/kotlin/com/bluebubbles/messaging/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/bluebubbles/messaging/MainActivity.kt
@@ -7,6 +7,7 @@ import androidx.activity.ComponentActivity
 import com.bluebubbles.messaging.services.backend_ui_interop.MethodCallHandler
 import com.bluebubbles.messaging.services.foreground.ForegroundServiceBroadcastReceiver
 import com.bluebubbles.messaging.Constants
+import com.bluebubbles.messaging.utils.NotificationUtils
 import io.flutter.embedding.android.FlutterFragmentActivity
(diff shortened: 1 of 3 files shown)


zlshames:
So if you have no way of testing it, and the code was all written for AI, do you know if it even compiles or runs? We already have a method channel for creating notification channels, and it's handled in the kotlin code:

https://github.com/BlueBubblesApp/bluebubbles-app/blob/547c7376f449683863e9f20b539009617319bbfe/android/app/src/main/kotlin/com/bluebubbles/messaging/services/notifications/NotificationChannelHandler.kt#L13-L54

I read over the code changes, and I still don't really understand how this would fix background notifications. If it's failing silently, do you have an Android Debug Log?

PR #2951, opened 2025-10-14
Add option to hide title bar on linux

elliotnash:
This PR adds the ability to complete hide the title bar on linux. This replaces the Use Custom TitleBar option with a TitleBar Style selection with the following options:

<table>
  <tr>
    <td>Native</td>
    <td>Custom</td>
    <td>Hidden</td>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
    <td>image</td>
   </tr> 
</table>


jjoelj:
~~Hey, can you change the base branch to `joel/desktop/1.15/1.15.5.0`~~ Nvm I can do that

Also, what's the use case for this?

wiggleforlife:
> Also, what's the use case for this?  

Looks like they're on a tiling window manager, where no apps have title bars. Instead, keyboard shortcuts are used to manage windows. BlueBubbles would most likely be the only app that still has a bar in this scenario.

elliotnash:
Sorry I missed this! 

Yeah, the use case is tiling window managers where application titlebars are hidden and the compositor provides its own minimal decorations. In these setups, the compositor can advertise a preference for server‑side decorations and inform clients that they are tiled, but Flutter/bitsdojo_window currently forces client‑side decorations in native mode and does not obey that hint.

For example, on my system all window control buttons are disabled; with useCustomTitlebar off, bitsdojo_window still reserves and draws an empty titlebar area above the app content (the left image). A "hidden" option for the titlebar lets the app surface be fully decoration‑less so the compositor’s own decorations can be the only visible ones, which is the expected behavior on tiling WMs.

lib/database/global/settings.dart
```diff
@@ -643,7 +643,7 @@ class Settings {
     s.pinRowsLandscape.value = map['pinRowsLandscape'] ?? 1;
     s.pinColumnsLandscape.value = map['pinColumnsLandscape'] ?? 4;
     s.maxAvatarsInGroupWidget.value = map['maxAvatarsInGroupWidget'] ?? 4;
-    s.useCustomTitleBar.value = map['useCustomTitleBar'] ?? true;
+    s.titleBarStyle.value = map['titleBarStyle'] != null ? BBTitleBarStyle.values[map['titleBarStyle']] : BBTitleBarStyle.custom;
```
jjoelj: Could you add a migration here from `useCustomTitleBar` to `titleBarStyle` similar to what we do with `useDesktopAccent` (when we changed the name from `useWindowsAccent`)? Otherwise, the update will change the settings for the user in an unexpected way.

jjoelj (changes requested):
Minor UX thing

PR #2954, opened 2025-11-12
Trying to Fix Windows since it drops and effectively doesn't work

4o4R:
lib/services/network/socket_service.dart (lines 34-196) now tracks connectivity state, drops the localhost override only when Wi‑Fi/Ethernet vanish, and immediately reconnects when Windows reports the link is back. A shared _scheduleSocketRecovery helper also kicks in for both disconnected and error socket states so the client automatically refetches the Firebase URL and restarts the socket instead of staying offline.
lib/app/layouts/settings/pages/server/server_management_panel.dart (lines 519-551) lets “Manually Sync Messages” run even when the socket is down, updating the subtitle and surfacing a snackbar warning rather than blocking the incremental sync request. This keeps REST-based sync usable during desktop socket drops.


jjoelj:
I think this is fixed by b8cbb6ecb5e9c69c9504a02247a1cb4c7fa26b1d

4o4R:
Thank you!

Issue #2413, opened 2023-03-10
Properly show invisible ink messages


zlshames:

iT3g4n:
Any updates on this?

zlshames:
> Any updates on this?

Not yet, unfortunately

PR #2979, opened 2026-03-09
Make Invisible Ink Effect Display on Chat Open

ZachGarcia42:
## Problem: #2413 Messages sent with invisible ink show up without the invisible ink effect when opening a chat.

This PR makes the effect show up right away without needing to tap the "sent with invisible ink" message.

I'm very new to frontend and flutter, so feel free to leave suggestions on how to improve the change!


ZachGarcia42:
Sorry opened against master. See corrected one on development: #2980

PR #2980, opened 2026-03-09
Display invisible ink by default

ZachGarcia42:
## Problem: #2413 Messages sent with invisible ink show up without the invisible ink effect when opening a chat.

This PR makes the effect show up right away without needing to tap the "sent with invisible ink" message.

I'm very new to frontend and flutter, so feel free to leave suggestions on how to improve the change!


zlshames:
Hey thanks for the contribution! I know this will be a pain, but I'm in the middle basically rewriting most of the app. I'm nearly done. But that also means that development is currently behind but a lot. Any chance you can rebase with the zach/isolate-updates+desktop branch? I plan to merge into development but I just haven't yet

zlshames (approved):

PR #2982, opened 2026-04-01
Support arbitrary emoji reactions from newer iOS

### LINKED ISSUE #2823: Support for iOS 18 expanded tapback emoji?
As of iOS 18, iMessage allows you to tapback with any emoji, not just the previously pre-defined 6 emoji. Bluebubbles doesn't support this and I haven't seen any indication that this is planned. Is it planned but I just missed it?

Thanks!

### LINKED ISSUE #2829: Multi-Emoji Tapbacks
With iOS 18 and MacOS Sequoia apple allows for tapbacks with all emojis instead of the standard selection(HaHa, !!, Thumbs up, Heart, etc) in iOS 17 and below. 

Now when a tapback uses a emoji's out of the iOS 17 selections, bluebubbles indicates this reaction as 'null' in the message preview and doesn't show the reaction in the app.
image

kilabyte:
## Summary

Newer iOS versions allow reacting with any emoji, not just the 6 classic tapbacks. The `associatedMessageType` for these contains the raw emoji character. Previously these were filtered out or rendered as null.

Closes #2829
Closes #2823

- Add `isValidReaction`/`isEmojiReaction` helpers to `ReactionTypes` with proper Unicode emoji validation
- Update all reaction filter points to recognize emoji reactions
- Render emoji reactions as text instead of SVG on iOS skin
- Fix notification text for emoji reactions
- Fix dedup logic for emoji reactions in `normalizedAssociatedMessages`
- Null-safe `getReactionEmoji` and `getReactionVerb` to prevent force-unwrap crashes

## Changes

- `lib/helpers/ui/reaction_helpers.dart` — New helpers: `_emojiRegex`, `_looksLikeEmoji`, `_stripRemovalPrefix`, `isValidReaction`, `isEmojiReaction`, `getReactionEmoji`, `getReactionVerb`
- `lib/app/layouts/conversation_view/widgets/message/reaction/reaction.dart` — iOS skin renders emoji reactions as Text instead of SVG
- `lib/app/layouts/conversation_view/widgets/message/popup/message_popup.dart` — Reaction popup uses new helpers, null-safe emoji lookup
- `lib/app/layouts/conversation_view/widgets/message/message_holder.dart` — Reaction filtering uses `isValidReaction`
- `lib/database/html/chat.dart`, `lib/database/io/chat.dart` — Notification suppression uses `isValidReaction`
- `lib/database/html/message.dart`, `lib/database/io/message.dart` — Reactions getter uses `isValidReaction`
- `lib/helpers/types/helpers/message_helper.dart` — Notification text and dedup logic updated

## Test plan

- [ ] Receive a classic tapback (love, like, dislike, laugh, emphasize, question) and verify it renders correctly
- [ ] Receive an emoji reaction (e.g. thumbs up emoji, heart emoji, custom emoji) from a newer iOS device and verify it renders as the emoji character
- [ ] Verify reaction removal notifications show correct text
- [ ] Verify notification suppression works for both classic tapbacks and emoji reactions when "notify reactions" is off
- [ ] Test across all default themes and all skins (iOS, Material, Samsung)

### DIFF
--- a/lib/app/layouts/conversation_view/widgets/message/message_holder/message_holder_reactions.dart
+++ b/lib/app/layouts/conversation_view/widgets/message/message_holder/message_holder_reactions.dart
@@ -31,7 +31,7 @@ class ReactionObserver extends StatelessWidget {
       final isFromMe = state.isFromMe.value;
       final associatedMessages = state.associatedMessages;
       final reactions = associatedMessages
-          .where((e) => ReactionTypes.toList().contains(e.associatedMessageType?.replaceAll("-", "")))
+          .where((e) => ReactionTypes.isValidReaction(e.associatedMessageType))
           .toList();
       final reactionList = messageParts.length == 1 ? reactions : reactionsForPart(part.part, reactions).toList();
       return Positioned(
@@ -96,7 +96,7 @@ class ReactionSpacing extends StatelessWidget {
       // Directly observe MessageState associatedMessages for reactivity
       final associatedMessages = state.associatedMessages;
       final reactions = associatedMessages
-          .where((e) => ReactionTypes.toList().contains(e.associatedMessageType?.replaceAll("-", "")))
+          .where((e) => ReactionTypes.isValidReaction(e.associatedMessageType))
           .cast<Message>()
           .toList();
       if ((messageParts.length == 1 && reactions.isNotEmpty) || reactionsForPart(part.part, reactions).isNotEmpty) {
--- a/lib/app/layouts/conversation_view/widgets/message/message_holder/message_holder_timestamps.dart
+++ b/lib/app/layouts/conversation_view/widgets/message/message_holder/message_holder_timestamps.dart
@@ -32,7 +32,7 @@ class SamsungTimestampObserver extends StatelessWidget {
       final isFromMe = ms.isFromMe.value;
       final associatedMessages = ms.associatedMessages;
       final reactions = associatedMessages
-          .where((e) => ReactionTypes.toList().contains(e.associatedMessageType?.replaceAll("-", "")))
+          .where((e) => ReactionTypes.isValidReaction(e.associatedMessageType))
           .toList();
       return Padding(
         padding: (messageParts.length == 1 && reactions.isNotEmpty) || reactionsForPart(part.part, reactions).isNotEmpty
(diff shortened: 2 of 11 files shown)


zlshames:
Hey thanks for the contribution! I know this will be a pain, but I'm in the middle basically rewriting most of the app. I'm nearly done. But that also means that development is currently behind but a lot. Any chance you can rebase with the `zach/isolate-updates+desktop` branch? I plan to merge into development but I just haven't yet

kilabyte:
Thanks for the quick response! No problem at all. I'll rebase onto `zach/isolate-updates+desktop` and update the PR. Will push the changes shortly.

zlshames:
Hey @kilabyte just to make sure, this is just for displaying those reactions right? It doesnt allow sending them to the server yet (since it doesn't support that), right? Just confirming, but that's what it looks like

kilabyte:
Hey @zlshames, yep that's correct! This is purely for displaying and handling incoming emoji reactions. No changes to the send flow or server communication. The reaction picker in the popup still only shows the 6 classic tapbacks.

zlshames (approved):

zlshames:
thank ya, looks good. Pulling to test out.

zlshames:
I've reverted the code for now. I don't think it was implemented properly. Your code is checking the `associatedMessageType` field, but on newer versions that support those emojis, it's part of a new column: `associated_message_emoji` (on the mac).

I know this was AI-generated, so just be careful with what you PR... Even if you do review it. Make sure to test it.

kilabyte:
ah ok i can make the fix. it was ai-tandem yes, not completely AI generated

kilabyte:
then looking deeper my quick fix to client for reading the emojis wont work and would need server side changes as well. since youre saying the associated_meesage_emoji is where its stored on the macs iMessage DB and the server never picks it up or sends it to the client. So theres no data for the client to display no matter what we do on the client side or are you planning to make that change as well?

kilabyte:
Just testing new changes now and running on my build server. Looks like there's already a pre-existing condition on this branch that the reaction pop-up does not come up.

zlshames:
> then looking deeper my quick fix to client for reading the emojis wont work and would need server side changes as well. since youre saying the associated_meesage_emoji is where its stored on the macs iMessage DB and the server never picks it up or sends it to the client. So theres no data for the client to display no matter what we do on the client side or are you planning to make that change as well?

Yeah, so this is a multi-part change. On macOS versions older than Sonoma, the reactions will just come in as regular messages like, "Reacted 💯 to 'xxxxxx'". On Sonoma and newer, there is the `associated_message_emoji` column in the DB, as well as (likely) some other flag telling us that it's an emoji reaction. Might be `associated_message_type` or some other field. Not sure yet. We need to add support for that on the server and _then_ we can update the app to support it based on the server's implementation of it. It will need to be tested on macOS Sonoma or newer to really know if it can work.

Issue #2751, opened 2024-06-17
New message received - last message still showing in message list


keithop:
I have the windows bluebubbles app open all the time.
I often receive "one time codes" from places I login to
I receive a notification from bb saying for example "your onetime code is 5678"
but when I look over to the bluebubbles app without clicking into the actual chat, the preview in the messages list still shows the last onetime code from that site "your onetime code is 1234"

In other words, the messages list doesnt show the LATEST message for that chat

bfeist:
Came here to write a similar issue. Bumping this.

Issue #2668, opened 2024-01-29
bluebubbles-server does not work when server URL includes a significant path after the port number


jxyzn:
I'd like to put BlueBubbles on my host with an existing SSL cert, letting it live as a sub-path of the main server, however the bluebubbles-server appears to assume it's always at the root, even when the "server URL" includes a subdirectory.

With lighttpd, I thought I could get this working with a lighttpd config section like this on "host.my.domain":

```
server.modules   += ( "mod_proxy" )
$HTTP["url"] =~ "^/bluebubbles/" {
   # WebSocket proxy
   proxy.server  = ( "" => ("internal.mac.ip.addr" => ( "host" => "internal.mac.ip.addr", "port" => 1234 )))
   proxy.replace-http-host = "enable"
   proxy.forwarded = ( "for" => 1 )
   proxy.header = (
      "https-remap" => "enable",
      "map-urlpath" => (
          "/bluebubbles" => "",
      ),
   )
}
```

This would let SSL-secured access to **https://host.my.domain:443/bluebubbles/*** be forwarded (minus the SSL) to **http://internal.mac.ip.addr:1234/** (with the same path info from the original URL up to "bluebubbles/" removed).

Setting my server URL to "**https://host.my.domain/bluebubbles/**" in the web app does not work: In trying this, I see the web app fails due to attempting to access **https://host.my.domain/api/v1/...etc...**, ignoring the path I specified. Additionally, changing the config to catch "**/api/v1/**" and pass those along, it fails with **/socket.io/** path URLs a little later. And adding that path to the patterns forwarded still has some websocket errors.

Is supporting a sub-path on the URL something that could be supported, or is there a structural problem with supporting that?

Thanks

Issue #2667, opened 2024-02-05
Messages out of order


cstraw2021:
Even when the timestamp updates, certain chats dont move to the top no matter how many messages come in
<img width="84" alt="Screenshot 2024-02-04 232638" src="https://github.com/BlueBubblesApp/bluebubbles-app/assets/86491289/dfc1c69a-d96c-4e18-8bfe-5b4f091b957f">

tneotia:
Can you specify which platform you experience this on

Issue #2656, opened 2024-01-23
Unable to type in windows app after clicking on notification


sah4jpatel:
Been having an issue for a little while where when I click on a windows bluebubbles notification, the text box does not register input and I usually have to back out to the chat list and then open up that chat again.

I am on v1.12.7 on Windows 11 23H2 using the app using the .exe installer (not windows store) with the app window aspect ratio being closer to candy bar iPhone shape.

Minor inconvenience for sure, but would be good to have addressed (even if its user error) eventually.

sp1nkick:
I'm seeing this too

sah4jpatel:
An Additional detail: the text is there that is type but it is not visible. I have to back out to the chat list and then back into the chat and then it is visible.

sah4jpatel:
Ill give it a shot rn

jjoelj:
Did that resolve the issue?

sah4jpatel:
I don't think so. Started experiencing it again on a fresh install where when the window is open and then I click out into a different window to do something else and then click back into the bluebubbles text box, it doesn't register key strokes until I keep clicking (rarely works) or just back out into the chat list and the go back. Happy to keep testing, its not as common an issue anymore, but still happens occasionally.

sah4jpatel:
The issue now mainly and consistently occurs when the taskbar icon is clicked to either bring the app from the background to the front, or to unminimize it. Slightly different behavior when minimize to tray is enabled. The text cursor blinks but no text is registered usually, sometimes it appears 15 seconds later but its not usable until i go back to the chat list  and go into the chat. In the minimize to taskbar mode, the cursor doesnt blink and the textbox isnt clickable.

jjoelj:
@sah4jpatel can you confirm whether this issue is still present in the latest beta?

PR #3005, opened 2026-04-22
Development

cameronaaron:
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2424, https://github.com/BlueBubblesApp/bluebubbles-app/issues/2610, https://github.com/BlueBubblesApp/bluebubbles-app/issues/2751, https://github.com/BlueBubblesApp/bluebubbles-app/issues/2667, https://github.com/BlueBubblesApp/bluebubbles-app/issues/3004

https://github.com/BlueBubblesApp/bluebubbles-app/issues/2754 — Tablet mode: selecting a chat didn't focus the text field. Fixed by adding requestFocus: false to the right-panel Navigator.
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2770 — Server URL was reverting to the Firebase URL, overwriting manually-set static/Tailscale URLs. Fixed by making fetchNewUrl() return without saving, and only persisting after a confirmed connection.
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2553 — Voice memo pause button stayed stuck after playback completed. Fixed by always re-subscribing the onPlayerStateChanged listener with the current animController, regardless of whether the PlayerController was cached or new.
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2656/https://github.com/BlueBubblesApp/bluebubbles-app/issues/1968 — Desktop notifications / window restore
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2419 — Enter to send doesn't return focus
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2570 — Desktop audio timestamp stuck
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2598 — Conversation says "shared location" when stopping location share
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2668 — Server URL path stripping (reverse proxy sub-path support)
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2424 — Newly created chats don't show until restart
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2610 — New conversation shows phone number instead of name
https://github.com/BlueBubblesApp/bluebubbles-app/issues/2751 — Chat list shows stale last message after update
_repositionChat now calls _scheduleListVersionUpdate(immediate: immediate) in the currentIndex == -1 branch


cameronaaron:
@copilot apply changes based on the comments in [this thread](https://github.com/BlueBubblesApp/bluebubbles-app/pull/3005#pullrequestreview-4152030985)

zlshames:
Did you assign copilot to this ticket?  Tbh I didn't know others could request it

cameronaaron:
> Did you assign copilot to this ticket? Tbh I didn't know others could request it

Yep!

lib/app/layouts/conversation_view/widgets/message/interactive/interactive_holder.dart
zlshames, line 1: Ideally, this `highPerfMode` check should be part of the UrlPreview widget, so it can be managed within, rather than as a one-off

lib/services/backend/incoming_message_handler.dart
zlshames, line 1: Doesn't `latestMessage` represent a "cached" version of the latest message state, relative to the current Chat object? Should we make it `dbLatestMessage` so we fetch the latest message each time? Might utilize more resources, but would be more accurate?

zlshames (changes requested):
I've made some comments, in addition to Copilot. Mind addressing them?

cameronaaron:
> I've made some comments, in addition to Copilot. Mind addressing them?

Will do!

lib/app/layouts/chat_creator/widgets/search_contact_tile.dart
```diff
@@ -103,7 +103,15 @@ class _ContactAddressRow extends StatelessWidget {
...
-    final subtitle = (label != null && label!.isNotEmpty) ? '$address  •  $label' : address;
+    // When the display name is the same as the address (unnamed contact), avoid
+    // showing the number twice. Just show the label below if there is one (#2610).
+    final nameIsAddress = displayName.numericOnly() == address.numericOnly() && displayName.numericOnly().isNotEmpty;
+    final String? subtitleText;
+    if (nameIsAddress) {
+      subtitleText = (label != null && label!.isNotEmpty) ? label : null;
+    } else {
+      subtitleText = (label != null && label!.isNotEmpty) ? '$address  •  $label' : address;
+    }
```
tneotia, line 114: Can we take this out of the build method and put it into a getter or separate function. Just pedantic stuff

android/app/src/main/AndroidManifest.xml
```diff
@@ -226,6 +226,14 @@
...
                 <data android:scheme="imessage" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SENDTO" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:scheme="sms" />
+                <data android:scheme="smsto" />
+                <data android:scheme="mms" />
+                <data android:scheme="mmsto" />
+            </intent-filter>
```
tneotia, line 236: Not sure if we want to hijack SMS/MMS deep links without formally supporting SMS. I'll defer to @zlshames here.

lib/app/layouts/conversation_view/widgets/message/interactive/url_preview.dart
```diff
@@ -135,6 +135,22 @@ class _UrlPreviewState extends State<UrlPreview> with AutomaticKeepAliveClientMi
...
+        padding: const EdgeInsets.all(10),
+        child: Text(
+          displayUrl,
+          style: context.theme.textTheme.bodyMedium?.copyWith(
+            color: context.theme.colorScheme.primary,
+            decoration: TextDecoration.underline,
+          ),
+        ),
+      );
+    }
```
tneotia, line 152: I would go about this a different way... we should just show a plain old message bubble widget with the url instead of a preview widget. No code necessary in this file.

In the current state, this would not even allow the user to tap and open the link.

lib/app/layouts/conversation_view/widgets/text_field/text_field_component.dart
```diff
@@ -101,8 +103,25 @@ class TextFieldComponentState extends State<TextFieldComponent> {
         chat!.isIMessage));
   }
 
+  /// Trigger a send with the user-configured delay (respects the sendDelay setting).
+  /// Replaces direct sendMessage() calls from keyboard events so that the delay
+  /// is honoured when sending via Enter / numpadEnter / sendWithReturn.
+  void _sendWithDelay() {
```
tneotia, line 109: We would definitely want some sort of UI logic when implementing this to inform the user its the send delay setting, and allow the message to be user-cancelable.

lib/app/layouts/conversation_view/widgets/text_field/text_field_suffix.dart
```diff
@@ -240,21 +240,29 @@ class _RecordingButton extends StatelessWidget {
...
+                await audioRecorder.start(const RecordConfig(bitRate: 320000), path: temp.path);
+                return;
+              }
+              await recorderController!.record(
+                sampleRate: 44100,
+                bitRate: 320000,
+              );
+            } catch (e) {
+              controller!.showRecording.value = false;
+              final msg = e.toString().toLowerCase().contains('space') || e.toString().toLowerCase().contains('storage')
```
tneotia, line 261: Are there defined error classes within AudioRecorder or RecorderController? Looking at the string representation of the error is not ideal... but if we have to, it is what it is

tneotia (changes requested):
Haven't looked through it all yet, this is prelim feedback. You have a couple of different commits on here which is good, but it would make it a *lot* easier if each issue resolved had its own commit (or even better its own PR) just to go through and review and test. Sometimes hard to tell which files were edited for which issues, and then I can pull your individual commits and test each issue :)

PR #3020, opened 2026-05-08
fix: lifecycle crash when ChatsService isn't registered yet

ManavA:
## Problem

on a cold-start desktop launch, the OS sometimes fires a focus/resume lifecycle event before \`StartupTasks.init()\` has gotten to \`GetIt.I.registerSingleton<ChatsService>()\`. \`LifecycleService.open()\` then runs \`StartupTasks.onAppResume()\` which calls \`ChatsSvc\`, throwing:

\`\`\`
Failure during app initialization: Bad state: GetIt: Object/factory with type ChatsService is not registered inside GetIt.
#0  throwIfNot (package:get_it/get_it_impl.dart:14)
...
#5  ChatsSvc (package:bluebubbles/services/ui/chat/chats_service.dart:24)
#6  StartupTasks.onAppResume (package:bluebubbles/helpers/backend/startup_tasks.dart:529)
#7  LifecycleService.open (package:bluebubbles/services/backend/lifecycle/lifecycle_service.dart:115)
#8  LifecycleService.didChangeAppLifecycleState (...:74)
\`\`\`

reproducible on Linux desktop, every cold start. probably also reachable on Windows depending on how fast the toplevel window grabs focus relative to startup.

## Solution

guard the three \`ChatsSvc\` paths in \`LifecycleService\` on \`GetIt.I.isRegistered<ChatsService>()\`:

- \`open()\` — bail before \`StartupTasks.onAppResume()\` if services aren't up yet
- \`close()\` — capture a single \`chatsReady\` bool and gate every \`ChatsSvc\` call (the active-chat capture, \`setActiveToDead()\`, and the focus-unfocus block)
- the new android \`inactive\`/\`hidden\` \`setActiveToDead()\` block in \`didChangeAppLifecycleState\`

same pattern already used for \`GlobalIsolate\` further down in \`close()\`. \`startup_tasks\` sets the active state itself when \`onStartup\` finishes, so dropping these early lifecycle events doesn't lose state.

## Testing

- linux desktop, cold start: no more \`Bad state\` failure during init; window opens cleanly
- linux desktop, focus/blur cycles after startup completes: \`ChatsSvc.setActiveToAlive\`/\`...ToDead\` still fire as before
- normal pause/resume on android emulator: unchanged

### DIFF
--- a/lib/services/backend/lifecycle/lifecycle_service.dart
+++ b/lib/services/backend/lifecycle/lifecycle_service.dart
@@ -76,7 +76,7 @@ class LifecycleService with WidgetsBindingObserver {
       // "chat is active" notification guard. setActiveToDead() is idempotent
       // (equality-checked inside) and will run again in close() when paused fires.
       // Also handles Samsung One UI which emits `hidden` before `paused`.
-      if (Platform.isAndroid) {
+      if (Platform.isAndroid && GetIt.I.isRegistered<ChatsService>()) {
         ChatsSvc.setActiveToDead();
       }
     } else {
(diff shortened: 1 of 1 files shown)


ManavA:
Closing — upstream development has independently added the same `GetIt.I.isRegistered<ChatsService>()` guards that this PR was proposing (visible in the conflict diff against current development). The race I was patching is no longer reachable, so this PR is a no-op now. Thanks!

zlshames:
Yeah, thanks for contributing! But we hit the same issue you did, and we fixed it already. We fixed it through a few different modifications. It wasn't just a matter of changing one thing, but it was an amalgamation of changes to fix the issue

PR #3032, opened 2026-05-22
fix: incoming FaceTime overlay was un-dismissable

ManavA:
## Bug
When an incoming FaceTime call comes in, the overlay dialog gets stuck on screen — Accept opens a new dialog on top of the still-present overlay, Ignore does nothing visible. The user is trapped.

## Cause
Two compounding issues:

1. `showFaceTimeOverlay` captured `Get.rawRoute!` immediately after `showDialog(...)`. But `showDialog` pushes its route asynchronously, so at that moment `Get.rawRoute` is still the parent route, not the dialog. The wrong route gets stored in `faceTimeOverlays[callUuid]`.

2. `hideFaceTimeOverlay` calls `Get.removeRoute(...)`, which uses Get's `searchDelegate(null)` to find the navigator. When the dialog was pushed against `Get.context!` (the root nav), that delegate doesn't always match — `removeRoute` silently no-ops and the dialog stays.

## Fix
Switch the storage from `Map<String, Route>` to `Map<String, VoidCallback>`. The closure captures the dialog's own `BuildContext` from inside the `builder:` (where it's guaranteed to be the dialog's context) and calls `Navigator.of(dialogContext).pop()` directly. That always targets the dialog itself, regardless of which navigator delegate Get currently considers active.

Also call `hideFaceTimeOverlay(callUuid)` before `intents.answerFaceTime(...)` in the Accept handler so the overlay is gone before the next dialog appears.

## Verified
Tested locally — incoming FaceTime overlay now dismisses cleanly on both Accept and Ignore.

### DIFF
--- a/lib/helpers/ui/facetime_helpers.dart
+++ b/lib/helpers/ui/facetime_helpers.dart
@@ -12,16 +12,21 @@ import 'package:flutter/services.dart';
 import 'package:get/get.dart';
 import 'package:bluebubbles/services/backend/notifications/notifications_service.dart';
 
-Map<String, Route> faceTimeOverlays = {}; // Map from call uuid to overlay route
+/// Map from call uuid to a closure that pops the dialog using the dialog's
+/// OWN Navigator (captured inside the builder). Previously this stored a
+/// `Route` and called `Get.removeRoute`, but `Get.removeRoute` searches via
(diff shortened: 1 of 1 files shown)


ManavA:

PR #3033, opened 2026-05-22
fix: FaceTime 'Generating link…' dialog hangs forever if server slow

ManavA:
## Bug
After tapping Accept on an incoming FaceTime call, a 'Generating link for call...' progress dialog appears with no Cancel button, no `barrierDismissible`, and no timeout on the HTTP call. If the server is slow, unreachable, or the tunnel URL has rotated, the user is stuck staring at a spinner forever with no way out except killing the app.

## Fix
Three changes to `answerFaceTime`:
1. Add a **Cancel** button to the dialog
2. Set `barrierDismissible: true` so tap-outside also closes it
3. Wrap `http.answerFaceTime(callUuid)` with `.timeout(const Duration(seconds: 20))` so a hung server can't strand the dialog

Also added a `progressOpen` flag + `closeProgress()` helper to avoid double-popping (which would close the page underneath if the user already dismissed via Cancel/barrier).

### DIFF
--- a/lib/services/backend/java_dart_interop/intents_service.dart
+++ b/lib/services/backend/java_dart_interop/intents_service.dart
@@ -113,9 +113,25 @@ class IntentsService extends GetxService {
   }
 
   Future<void> answerFaceTime(String callUuid) async {
+    // Track whether the progress dialog is still on the stack so we never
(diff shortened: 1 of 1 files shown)


ManavA:

PR #3034, opened 2026-05-22
fix: Linux desktop killed on spawned-isolate ObjectBox open instead of attaching

ManavA:
## Bug
On Linux desktop, the app sometimes silently exits during startup. Logs show the main isolate successfully opens ObjectBox, then the spawned global isolate also tries to open the same store, ObjectBox throws 'another store is still open using the same path', and `_initDatabaseDesktop`'s catch unconditionally treats ANY exception as 'another instance is probably running' → `exit(0)`. The app dies immediately, no window ever appears.

This is the same case `_initDatabaseMobile` already handles correctly via `Store.attach` (lines 110-113). Desktop just didn't have the same branch.

## Fix
In the catch block, check for the 'another store is still open using the same path' string FIRST. If matched, it's a same-process race between isolates — call `Store.attach` like the mobile path does. Only fall through to the Linux 'another instance is running' `exit(0)` for genuinely OS-level instance conflicts.

## Repro
Build any post-3.10 desktop Flutter version that spawns isolates during startup on a Linux session. Without this fix, the app exits at the GlobalIsolate's first ObjectBox open attempt.

### DIFF
--- a/lib/database/database.dart
+++ b/lib/database/database.dart
@@ -132,6 +132,19 @@ class Database {
       Logger.info("Opening ObjectBox store from path: ${objectBoxDirectory.path}");
       store = await openStore(directory: objectBoxDirectory.path);
     } catch (e, s) {
+      // A spawned isolate (e.g. the global isolate, sync isolate) opening the
+      // store that the main isolate already holds throws "another store is
+      // still open using the same path". ObjectBox forbids two open() calls
+      // for the same path in one process — Store.attach is the supported way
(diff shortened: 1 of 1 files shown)


ManavA:
Closing as duplicate of #3021, which is the cleaner version of this fix (platform-agnostic catch + upfront `Store.isOpen()` check mirroring `_initDatabaseMobile`, targeting `development`). Apologies for the duplicate — I forgot #3021 was already open when I bundled this batch.

PR #3037, opened 2026-05-23
v2.0.0-b8 candidate: sync perf, chat list refresh, FaceTime fixes + outgoing Link

ManavA:
Bundle of five focused commits I've been carrying in my BluerBubbles fork. Each is independently useful; happy to split into separate PRs if that's easier to review.

### `perf: parallelize per-chat message fetch in full sync (concurrency 6)` *(existing fork commit)*
Sequential per-chat HTTP roundtrips were the bottleneck on large accounts (N-chat account = N serial HTTP calls). Switched the per-chat loop to a bounded concurrency pool. Roughly N× faster wall-clock for the message-fetch phase.

### `perf: bump sync concurrency to 12, defer group-icon fetch, suppress empty-batch log spam`
- Concurrency 6 → 12. Most servers handle the extra HTTP load fine; roughly halves wall-clock time for the message-fetch phase.
- Group `getIcon` was awaited per-chat in the message-sync loop. Moved to fire-and-forget so a slow icon endpoint doesn't bottleneck the whole chat. Failures now log instead of getting silently swallowed.
- Sync log was emitting `"Saving chunk of 0 message(s) for chat: X"` for every empty chat, flooding the live log on accounts with thousands of inactive SMS shells. Only emit when there are actually messages.

### `fix: conversation list tile stale when chat is already at top after new message`
`_repositionChat` short-circuits the `chatListVersion` bump when the chat's sorted position doesn't change — common for active chats and self-sends. The chat stays at index 0, but the conversation list never rebuilds, so the tile keeps showing the old subtitle/timestamp. Repro: send yourself a message → message appears in the detail view but the list tile doesn't update preview/time until the next sort-affecting event.

Fix: force `_scheduleListVersionUpdate(immediate: true)` at the end of `updateChatLatestMessage` regardless of repositioning outcome.

### `fix: incoming FaceTime overlay silently aborts when server payload omits is_audio`
`bool isAudio = data["is_audio"]` threw `_TypeError(null is not bool)` whenever the server payload arrived without that field. The error propagated up through the async chain and the entire overlay path silently aborted — user saw nothing for incoming calls. Wrapped the whole handler in try/catch with `Logger.error`, and made each field null-safe with sensible defaults. Also skips the overlay (falls back to notification) when there's no callUuid, since the Accept button would have nothing to act on.

### `feat: outgoing FaceTime via Apple Link, replace dead Google Duo path`
The Video Call action in chat details was calling `MethodChannelSvc.actions.googleDuo()`, but **Google killed Duo in 2022** and merged it into Meet — the call always failed with a "Google Duo may not be installed" snackbar on Android.

New behavior:
- **macOS/iOS**: `facetime:` / `facetime-audio:` URI as before (native FaceTime).
- **Android/Linux/Windows (video)**: bottom sheet that hits the server's existing `POST /api/v1/facetime/session` endpoint, gets a real FaceTime Link from the Mac's Private API (FaceTime → Create Link), and offers **Copy / Join in browser / Send to <chat>** as one-tap actions. The link works for anyone — Apple or not — they tap it, sign in, join from a browser.
- **Android (audio)**: `tel:` for the native dialer.
- Other platforms: clear snackbar explaining the constraint.

Adds `FaceTimeApi.createSession()` to wrap the existing server endpoint. Threads optional `Chat` through `showAddressPicker` so the bottom sheet can offer "Send to this chat" when invoked from chat details.

## Testing
- Sync perf: tested against a 4000-chat account; ~6× → ~12× speedup observed.
- Chat list refresh: send-to-self on both Cupertino and Material skins — tile preview and timestamp now update immediately.
- FaceTime overlay: server with intermittent `is_audio` field — overlay now reliably appears.
- FaceTime Link: tested on Android, link generation + send-to-chat round trip works. Browser join also works.

### DIFF
--- a/lib/app/layouts/conversation_details/dialogs/address_picker.dart
+++ b/lib/app/layouts/conversation_details/dialogs/address_picker.dart
@@ -1,41 +1,207 @@
+import 'dart:io' show Platform;
+
 import 'package:bluebubbles/helpers/helpers.dart';
 import 'package:bluebubbles/database/models.dart';
 import 'package:bluebubbles/services/services.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:get/get.dart';
 import 'package:permission_handler/permission_handler.dart';
 import 'package:url_launcher/url_launcher.dart';
 
-void launchIntent(bool video, String address) async {
-  if (address.contains("@")) {
-    launchUrl(Uri(scheme: "mailto", path: address));
-  } else if (await Permission.phone.request().isGranted) {
-    if (video) {
-      try {
-        await MethodChannelSvc.actions.googleDuo(number: address);
-      } catch (_) {
-        showSnackbar("Error", "Something went wrong, Google Duo may not be installed!");
-      }
-    } else {
-      launchUrl(Uri(scheme: "tel", path: address));
+/// Launch an outgoing call. Cross-platform behavior:
+///
+/// - macOS/iOS: `facetime:` / `facetime-audio:` opens native FaceTime.
+/// - Android/Linux/Windows (video): hit the BB server's `/facetime/session`
+///   endpoint to create a real FaceTime Link via the Mac, then offer to
+///   copy or send it in the current chat. Replaces the dead Google Duo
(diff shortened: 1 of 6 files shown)


ManavA:

PR #3039, opened 2026-05-25
fix: facetime overlay silent abort + outgoing call via apple link instead of dead duo

ManavA:
two facetime issues.

incoming overlay never appears when the server payload omits is_audio. handleIncomingFaceTimeCall does `bool isAudio = data["is_audio"]` which throws _TypeError on null and silently aborts the whole flow, so user sees nothing for a real incoming call. wrapped the handler in try/catch with logging and made the fields null safe. also fall back to a notification if there's no callUuid since the overlay's accept button would have nothing to act on.

outgoing video call from android still tries MethodChannelSvc.actions.googleDuo. google killed duo in 2022 (merged into meet) so it just snackbars "google duo may not be installed". on apple platforms use the facetime: and facetime-audio: schemes which the os handles natively. on android/linux/windows show a bottom sheet that hits the existing /api/v1/facetime/session endpoint, gets a real facetime link from the mac, and gives copy / join in browser / send to chat. anyone can tap a facetime link to join, apple or not. wires up createSession() on FaceTimeApi (server endpoint is already there).

### DIFF
--- a/lib/app/layouts/conversation_details/dialogs/address_picker.dart
+++ b/lib/app/layouts/conversation_details/dialogs/address_picker.dart
@@ -1,41 +1,182 @@
+import 'dart:io' show Platform;
+
 import 'package:bluebubbles/helpers/helpers.dart';
 import 'package:bluebubbles/database/models.dart';
 import 'package:bluebubbles/services/services.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:get/get.dart';
 import 'package:permission_handler/permission_handler.dart';
 import 'package:url_launcher/url_launcher.dart';
 
-void launchIntent(bool video, String address) async {
-  if (address.contains("@")) {
-    launchUrl(Uri(scheme: "mailto", path: address));
-  } else if (await Permission.phone.request().isGranted) {
-    if (video) {
-      try {
-        await MethodChannelSvc.actions.googleDuo(number: address);
-      } catch (_) {
-        showSnackbar("Error", "Something went wrong, Google Duo may not be installed!");
-      }
-    } else {
-      launchUrl(Uri(scheme: "tel", path: address));
+void launchIntent(bool video, String address, {BuildContext? context, Chat? chat}) async {
+  final bool isApple = !kIsWeb && (Platform.isMacOS || Platform.isIOS);
+  final bool isAndroid = !kIsWeb && Platform.isAndroid;
+
+  if (address.contains("@") && !video) {
+    await launchUrl(Uri(scheme: "mailto", path: address));
+    return;
+  }
+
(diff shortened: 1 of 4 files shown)


lib/app/layouts/conversation_details/dialogs/address_picker.dart
```diff
@@ -1,41 +1,182 @@
...
-    if (video) {
-      try {
-        await MethodChannelSvc.actions.googleDuo(number: address);
-      } catch (_) {
-        showSnackbar("Error", "Something went wrong, Google Duo may not be installed!");
-      }
-    } else {
-      launchUrl(Uri(scheme: "tel", path: address));
+void launchIntent(bool video, String address, {BuildContext? context, Chat? chat}) async {
+  final bool isApple = !kIsWeb && (Platform.isMacOS || Platform.isIOS);
```
jjoelj, line 14: This is never true for BlueBubbles

lib/app/layouts/conversation_details/dialogs/address_picker.dart
```diff
@@ -1,41 +1,182 @@
...
+    await launchUrl(Uri(scheme: "mailto", path: address));
+    return;
+  }
+
+  if (video) {
+    if (isApple) {
+      final ok = await launchUrl(Uri(scheme: "facetime", path: address));
+      if (!ok) showSnackbar("Error", "Couldn't launch FaceTime for $address");
+      return;
+    }
```
jjoelj, line 27: As above

lib/app/layouts/conversation_details/dialogs/address_picker.dart
```diff
@@ -1,41 +1,182 @@
...
+    if (context != null && context.mounted) {
```
jjoelj, line 28: `context?.mounted == true`

lib/app/layouts/conversation_details/dialogs/address_picker.dart
```diff
@@ -1,41 +1,182 @@
...
+      return;
+    }
+    showSnackbar("FaceTime unavailable", "Tap Video Call from a chat to get the FaceTime Link option.");
+    return;
+  }
+
+  if (isApple) {
+    await launchUrl(Uri(scheme: "facetime-audio", path: address));
+    return;
+  }
```
jjoelj, line 39: As above

jjoelj (changes requested):
BlueBubbles does not support MacOS or iOS. None of that code is necessary.

For Android, the video call button does open Google Meet. That was fixed in db7ba07b4129becb7b3320ee5a6a8e033504aff0

Since facetime support is experimental and unreliable, I doubt that we will make this the default functionality on Android. It should be gated behind an option (the default continuing to be Google Meet). On desktop platforms, the button shouldn't show up unless that option is enabled.

PR #3042, opened 2026-05-25
fix: conversation list tile stale when chat already at top after new message

ManavA:
send yourself a message and watch the detail view update while the chat list tile keeps the old subtitle and timestamp. happens because _repositionChat short circuits the chatListVersion bump when the chat is already in its sorted slot (which it is for the active chat, self sends, anything pinned at the top).

force a _scheduleListVersionUpdate(immediate: true) at the end of updateChatLatestMessage regardless of whether reposition moved anything. one line.


jjoelj:
I can't reproduce this issue the way you're describing

PR #3045, opened 2026-05-31
feat: cross-device clipboard sync via BB server

hazeltr0n:
## Summary

- Adds a `ClipboardSyncService` that polls the local clipboard every second and sends changes to the BlueBubbles server via a `clipboard-sync` socket event
- Receives `clipboard-sync` events from the server and writes the content to the local clipboard (iOS receives only — Universal Clipboard handles iPhone↔Mac)
- Adds a **Clipboard Sync** settings panel (Advanced section, desktop-only) with an enable/disable toggle

## How it works

The Mac server is the hub. It polls its own clipboard (which picks up iPhone changes via Apple Universal Clipboard) and broadcasts `clipboard-sync` to all connected clients. Windows and Mac clients also send their local clipboard changes back to the server. iOS receives only.

## Test plan

- [ ] Enable Clipboard Sync in Settings → Advanced → Clipboard Sync on a connected desktop client
- [ ] Copy text on Mac server machine — confirm it appears on the client clipboard
- [ ] Copy text on Windows/Mac client — confirm it appears on the server clipboard
- [ ] Disable the toggle — confirm polling stops and no more sync events are sent
- [ ] iOS: confirm no system clipboard-access banner appears (receive-only, no polling)

### DIFF
--- a/lib/app/layouts/settings/pages/advanced/clipboard_sync_panel.dart
+++ b/lib/app/layouts/settings/pages/advanced/clipboard_sync_panel.dart
@@ -1,0 +1,82 @@
+import 'package:bluebubbles/app/layouts/settings/widgets/settings_widgets.dart';
+import 'package:bluebubbles/helpers/helpers.dart';
+import 'package:bluebubbles/services/services.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:get/get.dart';
+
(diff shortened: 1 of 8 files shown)


zlshames:
why is this even needed? We aren't an iCloud replacement. We also probably do not want to be polling the clipboard every second, let alone every 500ms like you have on the server PR.

This functionality should not be in the BlueBubbles App IMO

Issue #2845, opened 2024-11-09
Cannot Open Android App


nep2ner:
After updating the Android app via Obtainium, the app opened and I toggled the background service notifications off then back on, went in & out of one chat (but didn't perform any actions there), then I got a notification that BlueBubbles kept crashing. Now, BlueBubbles will not open even if I force kill it and/or clear the cache. 

```
type: crash
osVersion: google/bluejay/bluejay:15/AP3A.241105.007/2024110700:user/release-keys
package: com.bluebubbles.messaging:20002069, targetSdk 34
process: com.bluebubbles.messaging
processUptime: 502 + 722 ms
installer: dev.imranr.obtainium

java.lang.IllegalArgumentException: URLDecoder: Illegal hex characters in escape (%) pattern : %s%
	at java.net.URLDecoder.decode(URLDecoder.java:214)
	at java.net.URLDecoder.decode(URLDecoder.java:143)
	at io.socket.global.Global.decodeURIComponent(Global.java:28)
	at io.socket.parseqs.ParseQS.decode(ParseQS.java:29)
	at io.socket.engineio.client.Socket.<init>(Socket.java:195)
	at io.socket.engineio.client.Socket.<init>(Socket.java:169)
	at io.socket.client.Manager$Engine.<init>(Manager.java:555)
	at io.socket.client.Manager$1.run(Manager.java:233)
	at io.socket.thread.EventThread.exec(EventThread.java:55)
	at io.socket.client.Manager.open(Manager.java:222)
	at io.socket.client.Manager.open(Manager.java:212)
	at io.socket.client.Socket$3.run(Socket.java:126)
	at io.socket.thread.EventThread$2.run(EventThread.java:80)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
	at java.lang.Thread.run(Thread.java:1012)
```

zlshames:
This is a known (& fixed, internally) bug and is caused by a % in your password. Update will be released soon. Apologies!

PR #3048, opened 2026-06-08
fix: prevent Android startup crash from malformed server URL (#2845)

richardtru:
## Summary

Fixes a permanent startup crash on Android (issue #2845). If the stored server URL contains bare `%` characters (e.g. `%s%`), socket.io's Java `URLDecoder` throws `IllegalArgumentException` on a background thread outside any try/catch in `SocketIOForegroundService`, crashing the process. Because the foreground service starts on every launch when `keepAppAlive` is enabled, the app then crashes on *every* startup and can only be recovered by clearing app data.

**Fix:** validate `serverUrl` with `java.net.URI()` (synchronous, no I/O) before handing it to `IO.socket()`; a `URISyntaxException` updates the notification to "missing server URL" and returns gracefully.

**Logging hardening (same code path):** the server URL is no longer written to logcat (connect + connect-error lines), and only custom-header *keys* are logged, never their values — both could leak credentials.

## Scope note

This PR was originally bundled with an ObjectBox store-corruption recovery path. Per @zlshames' question below, I've **pulled that out** — I haven't personally reproduced a store corruption (it was inferred from issue reports), so guarding against it is speculative. I'll revisit it in a **separate PR** only if I can produce a concrete repro. Rebased onto `development` to clear the conflict.

## Test plan

- [ ] **Malformed URL (#2845):** set `serverAddress` to `http://192.168.1.1:1234/%s%test`, launch → foreground service logs "Server URL stored in preferences is malformed", notification updates, no crash.
- [ ] **Normal startup:** valid server URL → connects as before.

### DIFF
--- a/android/app/src/main/kotlin/com/bluebubbles/messaging/services/foreground/SocketIOForegroundService.kt
+++ b/android/app/src/main/kotlin/com/bluebubbles/messaging/services/foreground/SocketIOForegroundService.kt
@@ -19,6 +19,7 @@ import com.bluebubbles.messaging.services.backend_ui_interop.DartWorkManager
 import com.bluebubbles.messaging.utils.PersistentLog
 import io.socket.client.IO
 import io.socket.client.Socket
+import java.net.URI
 import java.net.URISyntaxException
 import java.net.URLEncoder
 import org.json.JSONObject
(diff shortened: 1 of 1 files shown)


zlshames:
So, i potentially get the malformed URL guard. But what about the objectbox store corruption. Have you experienced a corruption? I've not in my many many years of using the app. Is it something we even need to check for?

richardtru:
Good question — honestly, no, I haven't reproduced an ObjectBox store corruption myself; that half was inferred from the crash reports in #2845 and related threads, not from a repro I can point to. You've been running it for years without hitting it, so guarding for it speculatively isn't the right call.

I've scoped this PR down to just the malformed-URL guard, which *is* concretely reproducible (bare `%` in the stored URL → `URLDecoder` throws on a background thread → permanent startup crash). I also rebased onto `development` to clear the conflict, and trimmed the diff so it no longer re-touches the `PersistentLog`/pref-key lines you'd already reverted.

If I can put together a real repro for the store-corruption case, I'll bring that back as its own PR. Thanks for the patience here.

zlshames (approved):

zlshames:
thank ya

PR #3052, opened 2026-06-13
fix(socket): stop Windows reachability poller from tearing down healthy sockets

johndkos:
## Problem

On Windows desktop, the client repeatedly loses its connection — most noticeably on a local network. Logs from an affected install show `Internet reconnected, restarting socket...` firing **hundreds of times** on a steady ~33s interval, with the socket disconnecting far more often from the LAN address than from the remote URL. The network never actually drops.

## Root cause

In `lib/services/network/socket_service.dart`, the Windows-only `InternetConnection` (internet_connection_checker_plus) poller probes `serverAddress` every few seconds with a 3s timeout. On a `disconnected` result it called `handleStatusUpdate(SocketState.error, null)` — forcing the state to `error` **even when the websocket was still connected**. The next successful probe then saw `error` and ran a full `restartSocket()`.

So a single slow/timed-out probe (easy to hit on a LAN with a 3s timeout) manufactured its own disconnect→restart cycle, independent of the real socket health. socket.io already has its own reconnection plus the `_handleReconnectFailed`/`_reconnectTimer` backoff, so the poller was both redundant and actively destructive here.

## Fix

- Only escalate to `SocketState.error` when the socket isn't actually connected (`socket?.connected != true && state.value != connected`).
- Only `restartSocket()` on recovery when the socket is genuinely down.
- Bump the probe timeout 3s → 5s to tolerate normal LAN jitter.

The poller's original intent (detecting genuine Windows disconnects, where `onConnectivityChanged` is unreliable) is preserved — socket.io's `onDisconnect`/`onConnectError` still handle real drops.

## Testing

- Confirmed `Socket.connected` is a public property in the pinned `socket_io_client` 3.1.4.
- Minimal, logic-only change; no new imports.
- Easy to validate at runtime: on an affected Windows install, `Internet reconnected, restarting socket...` should drop from hundreds of occurrences to near zero.

### DIFF
--- a/lib/services/network/socket_service.dart
+++ b/lib/services/network/socket_service.dart
@@ -159,7 +159,9 @@ class SocketService {
         customCheckOptions: [
           InternetCheckOption(
             uri: Uri.parse(serverAddress),
-            timeout: const Duration(seconds: 3),
+            // 3s is too aggressive for a busy LAN server — a single slow probe
+            // would falsely report "no internet". 5s tolerates normal jitter.
+            timeout: const Duration(seconds: 5),
             responseStatusFn: (_) => true,
           ),
         ],
@@ -170,8 +172,16 @@ class SocketService {
       internetConnectionListener = internetConnection!.onStatusChange.listen((InternetStatus status) {
         Logger.info("Internet status changed: $status");
         if (status == InternetStatus.disconnected) {
-          handleStatusUpdate(SocketState.error, null);
-        } else if (state.value == SocketState.error) {
+          // A single failed reachability probe must NOT tear down a live
+          // websocket. socket.io already detects real drops via
+          // onDisconnect/onConnectError; forcing SocketState.error here on every
+          // transient LAN/probe timeout manufactures a restart loop (the cause
+          // of repeated disconnects on the home network). Only escalate if the
+          // socket truly isn't connected.
+          if (socket?.connected != true && state.value != SocketState.connected) {
+            handleStatusUpdate(SocketState.error, null);
+          }
+        } else if (state.value == SocketState.error && socket?.connected != true) {
           Logger.info("Internet reconnected, restarting socket...");
           restartSocket();
         }


johndkos:
Verified locally on Windows against a live server. ✅

Built this branch (`flutter build windows --release`, Flutter 3.44.2 stable, VS 2022 Desktop C++ toolchain) and ran it on the same setup that was previously flapping. The reachability poller still fires the occasional transient blip, but the patched listener now **absorbs** it instead of tearing down a healthy socket.

**Before** (stock 2.0 desktop beta) — poller blip tears the socket down:
```
14:29:14  Internet status changed: InternetStatus.disconnected
14:29:24  Internet status changed: InternetStatus.connected
14:29:24  Disconnected from socket at http://192.168.x.x:xxxx
14:29:24  Socket connected successfully to http://192.168.x.x:xxxx
```

**After** (this branch) — identical blip, socket untouched:
```
14:40:47  Internet status changed: InternetStatus.disconnected
14:40:57  Internet status changed: InternetStatus.connected
          (no "Disconnected from socket", no restart)
```

Over a 4-minute window: `Internet reconnected, restarting socket` = **0** and `Disconnected from socket` = **0**, versus roughly one teardown every ~33s before. Connection stable since.

johndkos:
Heads-up for whoever merges this: the identical poller block lives on `development` too — same `InternetConnection` setup and `Internet reconnected, restarting socket` logic in `lib/services/network/socket_service.dart`. It'll need the same change there, otherwise the fix can regress when `development` and the 2.0 desktop line get merged. Happy to open a matching PR against `development` if that's preferred.

jjoelj:
This negates the point of the listener. If your internet actually disconnects, the socket will take ~3 minutes to trigger the disconnect event, because of the pingInterval configuration on the server.

The purpose of the connection checker *is* to disconnect the socket when the connection drops. The real fix requires a server update.

Not sure why your setup can't handle this connectivity check. If it still disconnects with a 5 second timeout for you, that may be an issue with your environment.

johndkos:
Thanks for the context — you're right, and I'm closing this.

I hadn't accounted for the `pingInterval` making socket.io's native disconnect take ~3 minutes. Gating the listener behind `socket.connected` just trades the flapping for a multi-minute blind spot on a genuine drop, which is clearly worse — the checker exists precisely to avoid that. My mistake.

Digging into my own setup confirms your "it's the environment" read: the bare origin responds in 6–130 ms when the server is idle, but my Mac server stalls ~6 s servicing `/server/statistics/totals` (and the Private API is throwing `Selector not found`), and the 5 s reachability probe times out *during those stalls*. So the checker is doing its job — my server is the bottleneck. I'll chase the server update / Private API on my end instead.

Appreciate you taking the time to explain it.

PR #3061, opened 2026-06-30
fix: notification sender bugs + opt-in starred-contact DND bypass

blue-archon:
## Summary

This PR contains two bug fixes and one new feature, each in a self-contained commit. Happy for these to be taken individually or together.

### Bug fixes

**Unknown sender in group chat notifications** — A second message arriving rapidly in a group chat showed "Unknown" as the sender name. Root cause: the handle lookup on `Message` fell through to a DB query that raced with ObjectBox relation loading. Fix: check the in-memory transient `handle` field (populated by `addMessageToChat` before the DB put) before issuing the query. Also adds a `deep_map_normalize` utility to fix `LinkedHashMap<Object,Object>` type corruption on nested server payload maps after cross-isolate transfer (affects both FCM and UnifiedPush background paths).

**Notification Like/reaction button silently failing** — The `flutter.` key prefix was missing from `hasServerUrl()` and `hasAuthKey()` in `SettingsHelper.kt`, so `HttpService` always saw empty values and never made the HTTP call. Fixed in the feat commit below alongside the SettingsHelper changes for the DND feature.

### New feature: opt-in starred-contact DND bypass

Removes `setBypassDnd(true)` from the global notification channel (which bypassed DND for all contacts) and replaces it with per-notification logic: if the sender is starred in the Android contacts app, the notification `Person` is built with `setUri(lookupUri)` + `setImportant(true)`, letting Android's DND system grant the exception contact-by-contact.

The feature is **off by default**, controlled by a new toggle in Settings → Application Settings → Notification Settings → *Override DND for Favorites*.

**Implementation note on settings storage:** The toggle value (`dndFavoritesOverride`) is passed from Dart through the `create-incoming-message-notification` MethodChannel call rather than read from `SharedPreferences` on the Kotlin side. This avoids a storage mismatch introduced by `shared_preferences` v2.5.x: after the DataStore migration, new keys are written to Android DataStore, which is not accessible via the legacy `getSharedPreferences()` API used by `SettingsHelper`. Existing keys (written before migration) are unaffected.

**Conflict resolution note:** `DartWorkManager.kt` had a minor conflict with the upstream `APPEND_OR_REPLACE` enqueue fix. Resolved by combining both: `enqueueUniqueWork` with `APPEND_OR_REPLACE` inside the try-catch, so sequencing and error handling are both preserved.

## Commits

- `fix: resolve unknown sender in group notifications and infinite loading spinner`
- `feat: add starred-contact DND bypass with user-controlled settings toggle`
- `docs: add fork README documenting changes from v2.0.0+85`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,176 @@
+# BlueBubbles — Community Fork (v2.0.0+85)
+
+> **This is an unofficial community fork** of [BlueBubblesApp/bluebubbles-app](https://github.com/BlueBubblesApp/bluebubbles-app) based on **v2.0.0+85 (beta 8)**. It adds two feature tracks and fixes several bugs found in that release. It is not affiliated with or endorsed by the BlueBubbles project.
+
+---
+
+## What This Fork Adds
+
+### Feature 1: Per-Contact Do Not Disturb Bypass via Android Starred Contacts
+
+The stock app adds the entire BlueBubbles notification channel to the DND override list — meaning *every* incoming message bypasses Do Not Disturb, which defeats the purpose of DND entirely.
+
+This fork changes that. Instead of granting the whole app DND immunity, it uses Android's native **"Starred contacts can interrupt"** DND rule. Contacts you have marked as **Favorites** in your Android contacts app (the star in Google Contacts, the Favorites list in Samsung Contacts) can ring through DND. Everyone else respects it.
+
+**How it works (technical summary):**
+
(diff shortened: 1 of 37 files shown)


jjoelj:
Only PR relevant files. Breaking into multiple PRs will help us review.

blue-archon:
Thanks for the feedback. Split into two focused PRs and dropped the fork-specific README:

- Bug fixes: #3062
- DND feature: #3063

Closing this one.

PR #3063, opened 2026-06-30
feat: add opt-in starred-contact DND bypass with settings toggle

blue-archon:
## Summary

Replaces the global `setBypassDnd(true)` on the notification channel (which bypassed DND for all contacts) with per-notification logic: if the message sender is starred in the Android contacts app, the notification `Person` is built with `setUri(lookupUri)` + `setImportant(true)`, letting Android's DND system grant the exception contact-by-contact.

The feature is **off by default**, controlled by a new toggle:
> Settings → Application Settings → Notification Settings → **Override DND for Favorites**

## How it works

1. `nativeContactId` is threaded from the Dart notification pipeline through `MethodChannelActions`, `DartWorker`, and `UnifiedPushReceiver` down to `CreateIncomingMessageNotification`
2. New `ContactNotificationHelper` resolves the contact lookup URI and starred status from `nativeContactId`
3. `PushShareTargetsHandler` reuses the same `ContactInfo` to avoid a second contacts DB query for conversation shortcuts
4. The `setBypassDnd(true)` call is removed from `NotificationChannelHandler` — the channel itself no longer forces through DND

## Implementation note: settings storage

The toggle value is passed from Dart through the `create-incoming-message-notification` MethodChannel call as `dnd_favorites_override` rather than read from `SharedPreferences` on the Kotlin side. This avoids a storage mismatch introduced by `shared_preferences` v2.5.x: after the DataStore migration, new keys are only in Android DataStore, which is not accessible via the legacy `getSharedPreferences()` API used by `SettingsHelper`. Existing keys written before migration are unaffected.

Also fixes the `flutter.` key prefix missing from `hasServerUrl()` and `hasAuthKey()` in `SettingsHelper.kt`, which was causing the notification Like/reaction button to silently fail to send.

## What changed

- `android/app/src/main/kotlin/.../utils/ContactNotificationHelper.kt` — new file
- `android/app/src/main/kotlin/.../utils/SettingsHelper.kt` — prefix fix + no new settings reads
- `android/app/src/main/kotlin/.../services/notifications/NotificationChannelHandler.kt` — remove bypassDnd
- `android/app/src/main/kotlin/.../services/notifications/CreateIncomingMessageNotification.kt` — conditional DND via method channel arg
- `android/app/src/main/kotlin/.../services/system/PushShareTargetsHandler.kt` — ContactInfo overload
- `android/app/src/main/kotlin/.../services/system/OpenConversationNotificationSettingsHandler.kt`
- `android/app/src/main/kotlin/.../services/backend_ui_interop/DartWorkManager.kt`, `DartWorker.kt`
- `android/app/src/main/kotlin/.../UnifiedPushReceiver.kt`
- `lib/database/global/settings.dart` — `dndFavoritesOverride` RxBool field
- `lib/services/backend/notifications/notifications_service.dart`
- `lib/services/network/method_channel_actions.dart`
- `lib/services/backend/java_dart_interop/method_channel_handlers.dart`, `method_channel_service.dart`
- `lib/helpers/backend/startup_tasks.dart`
- `lib/app/layouts/settings/pages/system/notification_panel.dart` — toggle UI
- `lib/app/layouts/settings/widgets/search/settings_items_list.dart` — search tag

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -128,7 +128,7 @@ android {
             productFlavors.tanay.signingConfig signingConfigs.debug
             productFlavors.alpha.signingConfig signingConfigs.release
             productFlavors.beta.signingConfig signingConfigs.release
-            productFlavors.prod.signingConfig signingConfigs.release
+            productFlavors.prod.signingConfig signingConfigs.debug
             minifyEnabled false
             shrinkResources false
         }
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -28,6 +28,12 @@ if (flutterVersionName == null) {
     flutterVersionName = '1.0'
 }
 
+// firebase-iid classes are now bundled inside firebase-messaging; exclude the
+// standalone artifact to prevent duplicate class errors from transitive deps.
+configurations.all {
+    exclude group: 'com.google.firebase', module: 'firebase-iid'
+}
+
 android {
     namespace "com.bluebubbles.messaging"
     compileSdk 36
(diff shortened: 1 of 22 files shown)


blue-archon:
Closing in favor of #3064 — a corrected version that stacks the DND feature on top of the bugfix it depends on (#3062), so it builds on its own. This branch was accidentally cut from `development` without that dependency underneath it. Thanks!

PR #3064, opened 2026-07-01
feat: add opt-in starred-contact DND bypass with settings toggle

### LINKED ISSUE #2493: Honor the DnD started contacts to send notifications during DnD time
https://discord.com/channels/640229510662455326/746560677333827584/1114998647638208673

blue-archon:
### What

Adds an opt-in "Override DND for Favorites" toggle (default off) in Notification Settings. When enabled, a message from a starred/favorite Android contact breaks through Do Not Disturb; non-favorites follow normal DND.

### How

Replaces the global `channel.setBypassDnd(true)` on the New Messages channel (and the per-conversation settings channel), which bypassed DND for every sender, with per-notification logic: when the sender resolves to a starred contact, the notification `Person` is built with `setUri(lookupUri)` plus `setImportant` so Android's own DND rules grant the exception contact by contact.

A new `ContactNotificationHelper` resolves the lookup URI and starred status. `nativeContactId` is resolved in Dart from the sender's `ContactV2` and passed with the toggle through the `create-incoming-message-notification` channel, so there is no extra native contacts query and no SharedPreferences dependency.

Stacked on #3062.


blue-archon:
Rebased onto current development and simplified: settings are now natively readable and the sender `Person` is already built in `CreateIncomingMessageNotification`, so I dropped the SharedPreferences workarounds from the earlier version. About 10 files instead of 21.

blue-archon:
Rebased onto current master. Two things worth flagging:
- Resolved the notification-panel conflict from the settings rewrite. The "Override DND for Favorites" toggle now sits inside the new `Obx`-wrapped `SettingsSection`.
- Unstacked this from #3062. They touch disjoint files, so this is now just the DND feature on its own and can be reviewed independently.

blue-archon:
Rebased onto v2.1.0, applies cleanly. This is the fork's headline feature, and I'd like to land it upstream so the fork can shrink. It uses Android's native starred-contact DND exception instead of putting the whole channel on the override list. Happy to adjust the approach if you'd like changes.

PR #3070, opened 2026-07-04
chore: de-fork desktop notifications onto flutter_local_notifications

blue-archon:
## Problem

Desktop notifications run on the BlueBubbles-maintained fork of `local_notifier`, while Android/iOS already use `flutter_local_notifications` (`# mobile only` in pubspec). That's two notification stacks, one of which is a fork that has to be carried indefinitely.

There was also a real bug in the old path: clicking a desktop notification only navigated to the chat if that chat was **already active** — clicking a notification for any other chat did nothing.

## Fix (commit 1)

Use `flutter_local_notifications` on desktop too, dropping the `local_notifier` fork:

- A small app-owned adapter (`lib/services/backend/notifications/desktop_notification.dart`) mirrors the old `LocalNotification` surface (title/body/actions/close, `onClick`/`onClickAction`/`onInput`/`onClose` callbacks), so call sites stay almost unchanged. It routes `NotificationResponse` (body click / action button / inline text input) back to the right callback.
- Click-to-open now routes through `IntentsSvc.openChat(...)` — opens the chat regardless of what was active (fixes the bug above).
- FaceTime notifications are persistent (ongoing) while ringing.
- Platform details wired for Linux (`FilePathLinuxIcon`, theme sound, actions), Windows (audio presets, `incomingCall` scenario, action/input elements), macOS (Darwin settings).

## Fix (commit 2)

The unified stack made a long-requested option trivial: a **"Notification Sound"** toggle in Desktop Settings (default on). Off = silent popup; on + custom sound file (existing feature) = custom sound; on + no file = system default. The custom-sound tile hides when sound is off.

## Notes for review

- `NotificationsService.init` sets a Windows `appName`/`appUserModelId`/GUID for the toast registration — maintainers may want to adjust those branding values.
- macOS action buttons need a registered `DarwinNotificationCategory` to render (follow-up); notification body click works without it.

## Testing

- Linux release build validated in daily use: new-message toasts, click-to-open (both from wide and single-pane layouts), action buttons, sound toggle in all three states.
- Windows/macOS compile paths updated but not runtime-tested here.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -33,7 +33,6 @@ import 'package:get/get.dart';
 import 'package:google_ml_kit/google_ml_kit.dart' hide Message;
 import 'package:intl/date_symbol_data_local.dart';
 import 'package:local_auth/local_auth.dart';
-import 'package:local_notifier/local_notifier.dart';
 import 'package:path/path.dart' as p;
 import 'package:permission_handler/permission_handler.dart';
 import 'package:screen_retriever/screen_retriever.dart';
@@ -542,9 +541,6 @@ class _HomeState extends State<Home> with WidgetsBindingObserver, TrayListener {
         } else {
           trayManager.addListener(this);
         }
-
-        /* ----- NOTIFICATIONS INITIALIZATION ----- */
-        await localNotifier.setup(appName: "BlueBubbles");
       }
 
       if (!SettingsSvc.settings.finishedSetup.value) {
--- a/lib/services/backend/notifications/desktop_notification.dart
+++ b/lib/services/backend/notifications/desktop_notification.dart
@@ -1,0 +1,214 @@
+import 'dart:async';
+
+import 'package:bluebubbles/utils/logger/logger.dart';
+import 'package:flutter_local_notifications/flutter_local_notifications.dart';
+
+enum _DesktopSound { defaultSound, sms, callLoop, silent }
+
+class _Callbacks {
+  _Callbacks({this.onOpen, this.onAction, this.onReply});
+
+  final FutureOr<void> Function()? onOpen;
+  final FutureOr<void> Function(int index)? onAction;
+  final FutureOr<void> Function(String text)? onReply;
+}
+
+/// Desktop (Linux + Windows) notifications on top of flutter_local_notifications.
+///
+/// flnp reports every tap/action through one global callback, so this owns an
+/// id -> callbacks registry and [handleResponse] fans each response back to the
+/// notification that raised it. Each `show*` returns the notification id for a
+/// later [cancel].
+class DesktopNotifications {
+  DesktopNotifications._();
+
+  static const String _replyInputId = 'reply';
+  static const String _openActionKey = 'open';
+
+  static FlutterLocalNotificationsPlugin? _plugin;
+  static final Map<int, _Callbacks> _callbacks = {};
+  static int _idCounter = 1000;
+
+  static void registerPlugin(FlutterLocalNotificationsPlugin plugin) => _plugin = plugin;
+
+  static void handleResponse(NotificationResponse response) {
+    final int? id = int.tryParse(response.payload ?? '');
+    if (id == null) return;
+    final _Callbacks? cb = _callbacks[id];
+    if (cb == null) return;
+
+    switch (response.notificationResponseType) {
+      case NotificationResponseType.selectedNotification:
+        cb.onOpen?.call();
+        break;
+      case NotificationResponseType.selectedNotificationAction:
+        final String? actionId = response.actionId;
+        final String? input = response.input;
+        if (actionId == _openActionKey) {
+          cb.onOpen?.call();
+        } else if (input != null && input.isNotEmpty) {
+          cb.onReply?.call(input);
+        } else {
+          final int? index = int.tryParse(actionId ?? '');
+          if (index != null) {
+            cb.onAction?.call(index);
+          } else {
+            cb.onOpen?.call();
+          }
+        }
+        break;
+    }
+  }
+
+  static Future<void> cancel(int id) async {
+    _callbacks.remove(id);
+    try {
+      await _plugin?.cancel(id: id);
+    } catch (e, s) {
+      Logger.error('Failed to cancel desktop notification', error: e, trace: s, tag: 'DesktopNotifications');
+    }
+  }
+
+  /// Simple title/body notification (alias deregistered, failed-to-send).
+  static Future<int?> showText({
+    required String title,
+    required String body,
+    FutureOr<void> Function()? onOpen,
+  }) {
+    return _show(title: title, body: body, onOpen: onOpen);
+  }
+
+  /// Incoming-message notification (avatar, actions, optional reply field).
+  static Future<int?> showMessage({
+    int? id,
+    String? imagePath,
+    required String title,
+    String? body,
+    String? attributionText,
+    List<String> actionLabels = const [],
+    bool replyInput = false,
+    bool silent = false,
+    FutureOr<void> Function()? onOpen,
+    FutureOr<void> Function(int index)? onAction,
+    FutureOr<void> Function(String text)? onReply,
+  }) {
+    return _show(
+      id: id,
+      imagePath: imagePath,
+      title: title,
+      body: attributionText != null ? '$attributionText$body' : body,
+      actionLabels: actionLabels,
+      replyInput: replyInput,
+      showOpenAction: true,
+      sound: silent ? _DesktopSound.silent : _DesktopSound.sms,
+      onOpen: onOpen,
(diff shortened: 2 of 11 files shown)


zlshames:
@jjoelj can you confirm if we can do this? I see you had some semi-recent updates to the forked version of the package. But is there a reason we still need it or does the official package support what we need now

lib/services/backend/notifications/desktop_notification.dart
jjoelj, line 1: We don't need to match the local_notifier API. This abstraction should at most just be a way to simplify calling the functions on Windows and Linux. Right now it's basically a double abstraction for no reason. We only use a few types of notifications

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -308,12 +329,9 @@ class NotificationsService {
...
 
-    toast.onClose = (reason) async {
-      if (reason == LocalNotificationCloseReason.timedOut && faceTimeOverlays.containsKey(callUuid)) {
-        await toast?.show();
-      }
-    };
-
+    // The notification is persistent (resident on Linux, incomingCall scenario
+    // on Windows), so it stays visible until answered/ignored/cleared without
+    // needing a re-show on timeout (flnp exposes no dismiss callback anyway).
```
jjoelj: Don't need this

lib/main.dart
```diff
@@ -533,8 +532,7 @@ class _HomeState extends State<Home> with WidgetsBindingObserver, TrayListener {
           trayManager.addListener(this);
         }
 
-        /* ----- NOTIFICATIONS INITIALIZATION ----- */
-        await localNotifier.setup(appName: "BlueBubbles");
+        // Desktop notifications are initialized in NotificationsService.init().
```
jjoelj: Remove

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -96,6 +96,26 @@ class NotificationsService {
         }
       }
     }
+
+    if (kIsDesktop && !headless) {
+      // flutter_local_notifications replaces local_notifier on desktop. One
+      // shared plugin plus a single response router (DesktopNotification) fans
+      // taps/actions back out to each notification's callbacks.
```
jjoelj: Unnecessary

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -96,6 +96,26 @@ class NotificationsService {
...
+            appUserModelId: 'com.bluebubbles.messaging',
+            // Stable GUID identifying the app to the Windows notification platform.
+            guid: 'c7e6c9a2-3b1f-4e8a-9d5c-2f7b6a4e1d90',
+          ),
+          macOS: const DarwinInitializationSettings(),
+        ),
+        onDidReceiveNotificationResponse: DesktopNotification.handleResponse,
+      );
+      DesktopNotification.registerPlugin(flnp);
+    }
```
jjoelj: Why not combine with the Android initialization?

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -96,6 +96,26 @@ class NotificationsService {
...
+      await flnp.initialize(
+        settings: InitializationSettings(
+          linux: const LinuxInitializationSettings(defaultActionName: 'Open'),
+          windows: const WindowsInitializationSettings(
+            appName: 'BlueBubbles',
+            appUserModelId: 'com.bluebubbles.messaging',
+            // Stable GUID identifying the app to the Windows notification platform.
+            guid: 'c7e6c9a2-3b1f-4e8a-9d5c-2f7b6a4e1d90',
+          ),
+          macOS: const DarwinInitializationSettings(),
```
jjoelj: All MacOS-related items should be removed

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -465,19 +487,30 @@ class NotificationsService {
...
+  void _attachToastHandlers(DesktopNotification toast, Chat chat, Message message, String avatarPath,
       List<String> actions, bool multipleMessages,
       {bool deleteFileOnClose = true}) {
     toast.onClick = () async {
       _cleanNotificationState(chat.guid);
-      await _openChat(chat);
       await windowManager.show();
+      // Canonical chat-open path (same as conversation tiles / mobile notif taps):
+      // seeds ChatState and navigates. The old _openChat only navigated when the
+      // chat was *already* active, so clicking a notification never opened it.
```
jjoelj: AI comments routinely mention old code. This should be in the PR description. We don't need this in the current code

lib/database/global/settings.dart
jjoelj, line 1: Make the settings addition a separate PR

lib/services/backend/notifications/desktop_notification.dart
```diff
@@ -0,0 +1,247 @@
...
+            ? WindowsNotificationAudio.silent()
+            : WindowsNotificationAudio.preset(sound: _windowsSound, shouldLoop: _loop),
+      );
+
+  // ----- macOS -----
+  DarwinNotificationDetails get _darwinDetails => DarwinNotificationDetails(
+        presentSound: soundOption != DesktopNotificationSoundOption.silent,
+        attachments: imagePath != null ? [DarwinNotificationAttachment(imagePath!)] : null,
+        interruptionLevel: persistent ? InterruptionLevel.critical : null,
+      );
```
jjoelj: Remove

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -96,6 +96,26 @@ class NotificationsService {
...
+    if (kIsDesktop && !headless) {
+      // flutter_local_notifications replaces local_notifier on desktop. One
+      // shared plugin plus a single response router (DesktopNotification) fans
+      // taps/actions back out to each notification's callbacks.
+      await flnp.initialize(
+        settings: InitializationSettings(
+          linux: const LinuxInitializationSettings(defaultActionName: 'Open'),
+          windows: const WindowsInitializationSettings(
+            appName: 'BlueBubbles',
+            appUserModelId: 'com.bluebubbles.messaging',
```
jjoelj: This id isn't really used on Windows. Maybe replace with BlueBubbles.BlueBubbles

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -96,6 +96,26 @@ class NotificationsService {
...
+            // Stable GUID identifying the app to the Windows notification platform.
+            guid: 'c7e6c9a2-3b1f-4e8a-9d5c-2f7b6a4e1d90',
```
jjoelj, line 84: Based on https://github.com/MaikuB/flutter_local_notifications/issues/2738#issuecomment-4091289709 this may need to match the `appUserModelId`

lib/services/backend/notifications/desktop_notification.dart
```diff
@@ -0,0 +1,247 @@
...
+    this.attributionText,
+    this.duration = DesktopNotificationDuration.short,
+    this.actions = const <DesktopNotificationAction>[],
+    this.hasInput = false,
+    this.inputPlaceholder,
+    this.inputButtonText,
+    this.systemSound,
+    this.soundOption,
+    this.persistent = false,
+    this.showOpenAction = false,
```
jjoelj: Rather than abstracting to a generic notification, it makes more sense to abstract to showing certain notifications. i.e. one for Facetime, one for texts. This entire file is far too abstracted

jjoelj (changes requested):
I'm a fan of consolidating to one package, but this implementation is abstracted too much. I've left a few initial comments.

The desktop sound setting seems useful, so put that in a separate PR it'll be merged faster.

blue-archon:
Reworked it. desktop_notification.dart is task-specific now (showMessage/showFaceTime/showText) with a small response router instead of the generic object. Pulled the macOS code, folded the desktop init into the Android initialize call, and dropped the extra comments plus the main.dart line. Renamed appUserModelId to BlueBubbles.BlueBubbles too, though I don't have a Windows box to confirm the guid/aumid bit from that issue, so a second look there would help. Split the sound setting off into its own PR, which I'll open once this lands since it sits on top of it.

zlshames:
@jjoelj when you have a chance, can you please re-review this?

jjoelj:
In its current state, much of this doesn't work correctly on Windows. The icon doesn't work, avatars don't show in the right place, there's an unnecessary 'open' action, causing notifications to fail to show with settings that worked before. @blue-archon if you can't test on Windows, I'll just implement this myself. The majority of desktop users are on Windows and Windows notification functionality is more complicated. I don't want to go back and forth with AI about it. Thanks for showing that the plugin supports desktop notifications now.

PR #3071, opened 2026-07-04
fix: show desktop notifications in the narrow single-pane window layout

blue-archon:
## Problem

`tryCreateNewMessageNotification` suppresses notifications when the user is "on the chat list" (`activeChat == null && route == "/" && !notifyOnChatList`). That's a mobile concept — on a phone, the list being foreground means the user is looking at it.

On desktop the same condition means "no chat pane is open", which is **always true in the narrow single-pane window layout**. Result: shrink the window horizontally and all message notifications are silently dropped (no toast, no sound). It also silences a wide window with no conversation selected.

## Fix

Gate the chat-list suppression to mobile (`!kIsDesktop`). Desktop now notifies regardless of window width or pane state — except for the chat the user is actively viewing, which is still suppressed by the existing `isChatActive` check above.

One-condition change (+6/−1) with a comment explaining why the rule is mobile-only.

## Testing

- Linux release build: narrow single-pane window on the chat list now produces notifications; messages for the actively-open chat stay silent; wide split-view behavior unchanged.

### DIFF
--- a/lib/app/layouts/settings/pages/system/notification_panel.dart
+++ b/lib/app/layouts/settings/pages/system/notification_panel.dart
@@ -51,7 +51,9 @@ class _NotificationPanelState extends State<NotificationPanel> with SingleTicker
                     child: Text("Notifications".psCapitalize, style: iOS ? iosSubtitle : materialSubtitle),
                   )),
             Obx(() => SettingsSection(backgroundColor: tileColor, children: [
-              if (!kIsWeb)
+              // On desktop this toggle has no effect: notifications there fire
+              // regardless of the chat list, so only show it on mobile.
+              if (!kIsWeb && !kIsDesktop)
                 SettingsSwitch(
                   onChanged: (bool val) async {
                     SettingsSvc.settings.notifyOnChatList.value = val;
@@ -77,7 +79,8 @@ class _NotificationPanelState extends State<NotificationPanel> with SingleTicker
                           : "Click to enable notifications",
                   backgroundColor: tileColor,
                 ),
-              const SettingsDivider(padding: EdgeInsets.only(left: 16.0)),
+              // Matches the chat-list toggle above: no leading divider on desktop where that toggle is hidden.
+              if (!kIsDesktop) const SettingsDivider(padding: EdgeInsets.only(left: 16.0)),
               SettingsSwitch(
                 onChanged: (bool val) async {
                   SettingsSvc.settings.notifyReactions.value = val;
--- a/lib/services/backend/notifications/notifications_service.dart
+++ b/lib/services/backend/notifications/notifications_service.dart
@@ -246,7 +246,9 @@ class NotificationsService {
     if (chat.shouldMuteNotification(message)) return;
     if (!headless && LifecycleSvc.isAlive) {
       if (ChatsSvc.isChatActive(chat.guid)) return;
-      if (ChatsSvc.activeChat == null &&
+      // Desktop always notifies for inactive chats; this chat-list suppression is mobile-only.
+      if (!kIsDesktop &&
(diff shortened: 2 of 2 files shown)


zlshames (commented):
@jjoelj can you confirm this? For desktop, I assume we want notifications so long as the notification is for a non-active chat or the window is not focused.

jjoelj:
Yeah, if the app is in the foreground and `notifyOnChatList` is disabled, notifications are not sent when no chat is active. I think the `notifyOnChatList` settings tile should be removed on desktop platforms since it does nothing with this change.

This doesn't really have to do with window layout at all, but the code is still useful.

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -210,7 +210,12 @@ class NotificationsService {
     if (chat.shouldMuteNotification(message)) return;
     if (!headless && LifecycleSvc.isAlive) {
       if (ChatsSvc.isChatActive(chat.guid)) return;
-      if (ChatsSvc.activeChat == null &&
+      // The "on the chat list" suppression is a mobile behaviour — there the list
+      // is the full foreground screen. On desktop, notifications are expected to
+      // fire regardless of window width or whether a chat pane is open, so a
+      // shrunk (single-pane) window shouldn't silence them. Mobile only.
```
jjoelj: This comment can be way more concise

lib/services/backend/notifications/notifications_service.dart
```diff
@@ -210,7 +210,12 @@ class NotificationsService {
...
+      if (!kIsDesktop &&
+          ChatsSvc.activeChat == null &&
           Get.rawRoute?.settings.name == "/" &&
           !SettingsSvc.settings.notifyOnChatList.value) {
```
jjoelj, line 253: the `notifyOnChatList` settings tile should be removed on desktop platforms since it does nothing with this change

jjoelj (changes requested):

blue-archon:
Hidden the "Send Notifications on Chat List" tile on desktop since it's a no-op there now, and dropped its divider so the section doesn't lead with one. Left the setting itself in place for mobile.

lib/app/layouts/settings/pages/system/notification_panel.dart
```diff
@@ -77,7 +79,7 @@ class _NotificationPanelState extends State<NotificationPanel> with SingleTicker
                           : "Click to enable notifications",
                   backgroundColor: tileColor,
                 ),
-              const SettingsDivider(padding: EdgeInsets.only(left: 16.0)),
+              if (!kIsDesktop) const SettingsDivider(padding: EdgeInsets.only(left: 16.0)),
```
jjoelj, line 83: Should probably match the previous condition here for consistency, so `!kIsWeb && !kIsDesktop`

jjoelj (changes requested):

blue-archon:
Done: divider now uses `!kIsWeb && !kIsDesktop` to match the tile above it, and trimmed the comment down to one line. The tile removal was already in the previous push.

blue-archon:
Rebased onto current master. Note: the notification settings panel was rewritten upstream, and it now has a separate web-only permission tile between this toggle and the divider. So this guards the divider with `!kIsDesktop` instead of `!kIsWeb && !kIsDesktop`. That keeps the divider on web (where the permission tile still needs it) and only drops it on desktop. Can match them back if you'd prefer.

blue-archon:
Rebased onto v2.1.0. The review asks were addressed (hide the chat-list toggle on desktop, divider guard matched to the toggle). Ready for re-review.

jjoelj (approved):

PR #3072, opened 2026-07-04
chore: remove dead Tenor GIF picker (public API discontinued)

blue-archon:
## Problem

Google discontinued the public Tenor API: new API keys/integrations were blocked on **2026-01-13** and the API was fully sunset on **2026-06-30** (the shutdown that forced X/Discord/WhatsApp/Bluesky to migrate their GIF pickers). There is no way to obtain a key anymore, so the composer's GIF button is now a dead control for every user — and on builds without a `TENOR_API_KEY` in `.env` it throws silently on every tap (`dotenv.get` on a missing key).

## Fix

Remove the dead feature:

- GIF button removed from the text-field icon bar
- `tenor_flutter` dependency dropped (it was pinned for Flutter-SDK compat — recurring upgrade friction), along with its four now-orphaned transitive lock entries
- the `TENOR_API_KEY` placeholder file (`lib/database/io/tenor.dart`) and its barrel export deleted; orphaned imports pruned

Net **−147 lines**. Android is unaffected — it never showed this button (uses the native/keyboard GIF flow).

If a replacement picker (Giphy/Klipy, as other platforms chose) is ever wanted, it would be a fresh integration anyway; this PR just stops shipping a button that can't work.

## Testing

- Linux release build; `flutter analyze` clean; composer renders correctly without the button.

### DIFF
--- a/lib/app/layouts/conversation_view/widgets/text_field/text_field_icon_bar.dart
+++ b/lib/app/layouts/conversation_view/widgets/text_field/text_field_icon_bar.dart
@@ -1,10 +1,7 @@
-import 'dart:async';
-
 import 'package:bluebubbles/app/layouts/conversation_view/widgets/text_field/conversation_text_field_local_controller.dart';
 import 'package:bluebubbles/database/models.dart';
 import 'package:bluebubbles/helpers/helpers.dart';
 import 'package:bluebubbles/services/services.dart';
-import 'package:bluebubbles/utils/logger/logger.dart';
 import 'package:bluebubbles/utils/share.dart';
 import 'package:chunked_stream/chunked_stream.dart';
 import 'package:file_picker/file_picker.dart' as pf;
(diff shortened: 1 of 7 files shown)


zlshames (commented):
While this makes sense. I don't just want to remove the GIF picker. We should replace it with something, rather than nothing. I know short-term, removing it is fine, but i think we just need to comment the code out instead of just dropping it since the new API we use might just be a drop-in replacement for the most part.

lib/app/layouts/conversation_view/widgets/text_field/text_field_icon_bar.dart
```diff
@@ -130,99 +125,6 @@ class TextFieldIconBar extends StatelessWidget {
             },
           ),
         ),
-        if (!kIsWeb && !Platform.isAndroid)
```
zlshames, line 133: Instead of deleting this code, can we just comment it out for now so we have some code to go off of when we replace it with a different GIF API

zlshames (changes requested):

blue-archon:
Restored the picker implementation as a commented-out block, and makes sense to keep as a placeholder

zlshames (approved):

zlshames (approved):

PR #3074, opened 2026-07-04
fix: never silently drop an outbound message when the prep phase fails

blue-archon:
## Problem

`OutgoingMessageHandler.queue()` runs the prep phase (`_prepItem` → `prepMessage`/`prepAttachment`, which writes the temp message through `GlobalIsolate.send`) with **no error handling**. The only failure recovery — `_finalizeOutgoingFailure`, which marks the message failed and shows the red bubble — lives in the later dispatch phase (`_processNext`'s `catchError`).

`queue()` is called fire-and-forget from the UI, so if anything throws during prep, the message is **silently dropped**: never enqueued, never sent, never marked failed, no bubble, nothing in the chat. The user's only clue is noticing later that the conversation's last message isn't theirs.

Prep can throw whenever the global isolate is mid-restart (e.g. right after an idle/drain shutdown — see companion PR), and also for file I/O during `prepAttachment`, DB errors, etc. We hit this in the wild repeatedly on a fork: plain text sends vanishing after backgrounding/resuming the app, worst in group chats.

## Fix

Wrap the prep phase in `queue()`:

- **Retry transient failures with backoff — but only while the temp record is not yet in the DB.** `prepMessage` regenerates the temp GUID on every call, so re-running after a partial save would duplicate the message; the not-in-DB guard makes retry safe exactly in the save-phase-failure case (the common one).
- **After retries are exhausted, or if the temp was already saved, surface the failure** via `_finalizeOutgoingFailure` — failed/retryable bubble, plus the failed-to-send notification when backgrounded — instead of vanishing. (`_finalizeOutgoingFailure` already tolerates a temp that was never saved: its `replaceMessage` is try/caught.)
- The happy path is unchanged — prep succeeds → enqueue → dispatch, identical to before.

Net effect: **an outbound message can no longer disappear silently**; worst case it shows as failed and can be retried.

## Testing

Validated on-device (Android, daily use for several days on a fork carrying this patch): the previously-reproducible silent loss no longer occurs; normal sends unaffected. `flutter analyze` clean. Independent of, but complementary to, the drain-cancel PR — each applies on its own.

### DIFF
--- a/lib/services/backend/outgoing_message_handler.dart
+++ b/lib/services/backend/outgoing_message_handler.dart
@@ -20,6 +20,11 @@ import 'package:universal_io/io.dart';
 
 const _tag = 'OutgoingMessageHandler';
 
+/// Outgoing-prep retry policy: retry a transient prep failure up to
+/// [_maxPrepAttempts] times, backing off [_retryBackoffMs] ms times the attempt.
+const _maxPrepAttempts = 3;
+const _retryBackoffMs = 250;
+
 // ignore: non_constant_identifier_names
 OutgoingMessageHandler get OutgoingMsgHandler => GetIt.I<OutgoingMessageHandler>();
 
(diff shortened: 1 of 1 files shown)


zlshames (commented):
i think this might resolve the issue at the surface, but I can't help but feel like this isn't a great resolution to the underlying problem. If the underlying problem is that the `addMessage` call may fail due to issues with the GlobalIsolate (maybe that's draining it or some other issue), i think we need a better fix for this. Whether that's fixing the underlying GlobalIsolate issue or having a more targeted fix. For instance, not using the GlobalIsolate for adding the message to the chat. Or maybe some other fix. I just don't love the retries without resolving the underlying issue. I know there was a cancel drain PR that you opened. That looked good, and maybe that'll help resolve this issue partially.

blue-archon:
Right, I agree. After further testing the silent drop still occurred even with the drain fix in place, so #3075 covers that trigger but not other ones. Prep can also fail for other reasons, like the isolate restarting or file I/O during prepAttachment, so I think the main value here is that a failed send becomes visible to the user instead of silently vanishing. On the underlying issue, I've added a watchdog and enhanced logging in my environment to narrow down the root cause. Once I have more certainty I'll follow up, possibly with another PR.

blue-archon:
I think I've got it (haven't hit it for a few days) - assuming no setbacks, should have something tomorrow, as well as a couple of other PRs for latent issues.

blue-archon:
Following up - opened #3083 with the root-cause fix. Platform-channel calls made from inside background isolates stop completing around app-pause transitions and jam the isolate's single-threaded event loop, so one hung prefs write (or a hung init) freezes everything queued behind it. The fix hands the isolate its filesystem paths and a prefs snapshot at spawn so it makes zero channel calls, plus a watchdog that restarts a wedged isolate instead of letting it freeze the app. Validated over a multi-day period on desktop and Android with no recurrence.

A few other PRs opened as well.

zlshames:
Ok, i have a few things to mention and call out:

1. **Pull the retry+finalize logic out of `queue()` into its own private method** (e.g. `_prepItemWithRetry(item)`). This file's convention is small, single-purpose helpers — `_prepItem`, `_copyWithMessage`, `_finalizeOutgoingFailure`, `_matchMessageWithExisting`. Each doing one thing with a doc comment. Right now `queue()` interleaves "prep," "retry/backoff," and "failure finalization" all inline, which makes it the longest, most nested method in the file. Extracting it would let `queue()` read as a clean top-to-bottom "prep -> enqueue -> dispatch," and the retry mechanics would be independently readable/testable.
2. **Name the backoff constant.** `maxPrepAttempts` is named, but `250 * attempt` is a magic number sitting right next to it. Something like `const _retryBackoffMs = 250;` next to `maxPrepAttempts` (or both hoisted to class-level constants near `_tag` at the top of the file, since that's where similar constants live) would match the rest of the file's style.
3. **Drop the `guid ?? ''` fallback, or make it a real guard.** Every call site (`send_animation.dart`, `notifications_service.dart`, `messages_service.dart`, etc.) calls `message.generateTempGuid()` before invoking `queue()`, so `item.message.guid` should never actually be null here. As written, if it somehow were null, you'd silently pass an empty-string guid into `_finalizeOutgoingFailure`, which would fail its internal `Message.replaceMessage` lookup and get swallowed by that method's own try/catch (`outgoing_message_handler.dart:941`). So, the failure would never reach the UI, quietly recreating the exact "silent drop" bug this PR exists to fix. Either use `item.message.guid!` (matching the invariant) or, if there's a real reason to distrust it, log loudly rather than pass `''`.
4. **Multi-part URL-split edge case**: the "was it saved" check only inspects `item.message.guid`, which is just the *first* message when `prepMessage` splits a long URL into two (pre-Big Sur macOS). If message 1 saves but message 2 throws, message 1 gets incorrectly finalized as failed and message 2's failure isn't tracked at all. Worth at least a comment acknowledging the limitation, or checking all generated guids rather than just the item's original one.
5. **Minor style nit:** `for (int attempt = 1;; attempt++) { ... return; ... return; }` (unconditional-return infinite loop) is a slightly unusual shape for "retry up to N times." A `for (int attempt = 1; attempt <= maxPrepAttempts; attempt++)` with the give-up path after the loop (or a plain `while (true)`) would make the exit conditions a bit more obvious at a glance. This is purely a readability preference, not a correctness issue.

I think this mostly looks good. Just a few nitpicks to check out.

blue-archon:
All five addressed:
1. Pulled the retry + finalize loop out of `queue()` into `_prepItemWithRetry()`.
2. Named the constants (`_maxPrepAttempts`, `_retryBackoffMs`) up by `_tag`.
3. Dropped the `''` fallback — using `item.message.guid!` to match the invariant.
4. Commented the pre-Big-Sur URL-split limitation.
5. Bounded `for` loop with the give-up path after it, instead of the infinite loop.

Note for #3: `_prepItem` returns `null` on *success* for attachments, so the helper returns a small `({bool ok, dynamic result})` record rather than using `null` as the failure sentinel. Pushed.

PR #3078, opened 2026-07-08
fix: refresh conversation list when a new chat is added at runtime

blue-archon:
A chat from a new incoming number, or one made in the chat creator, gets inserted into _sortedChats but the list doesn't re-render until something else bumps chatListVersion, so it stays invisible until you tap another chat. addChat inserts via _insertChatSorted without scheduling a list-version update; the batch loader does that itself, so only single runtime adds were affected. Also added the updateChat -> addChat fallback in the incoming handler for a brand-new chat, matching what the chat creator already does.

Hit it on linux desktop, but the code is shared so it should be validated on android too.

### DIFF
--- a/lib/services/backend/incoming_message_handler.dart
+++ b/lib/services/backend/incoming_message_handler.dart
@@ -379,6 +379,10 @@ class IncomingMessageHandler {
         unawaited(ChatsSvc.setChatHasUnread(c, false, force: true));
       }
 
+      // Seed the latest message before the ChatState is built so a freshly-added
+      // chat's tile has its subtitle on first paint.
+      c.setLatestMessage(saved);
+
       if (!ChatsSvc.updateChat(c, override: true)) {
         await ChatsSvc.addChat(c, immediate: true);
       }
--- a/lib/services/backend/incoming_message_handler.dart
+++ b/lib/services/backend/incoming_message_handler.dart
@@ -379,7 +379,9 @@ class IncomingMessageHandler {
         unawaited(ChatsSvc.setChatHasUnread(c, false, force: true));
       }
 
-      ChatsSvc.updateChat(c, override: true);
+      if (!ChatsSvc.updateChat(c, override: true)) {
+        await ChatsSvc.addChat(c, immediate: true);
+      }
       ChatsSvc.updateChatLatestMessage(c.guid, saved);
     }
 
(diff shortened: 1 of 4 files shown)


lib/services/backend/incoming_message_handler.dart
```diff
@@ -379,7 +379,13 @@ class IncomingMessageHandler {
         unawaited(ChatsSvc.setChatHasUnread(c, false, force: true));
       }
 
-      ChatsSvc.updateChat(c, override: true);
+      // Seed the latest message before the ChatState is built so a freshly-added
+      // chat's tile has its subtitle on first paint.
+      c.setLatestMessage(saved);
```
zlshames: I don't think we want to do this here. It should be done when the message is synced to the DB, and that has guards around when to update it which i think i want to maintain. That's the only change I'd make.

blue-archon:
Yes, makes sense. The latest-message link now happens on the guarded sync path (`Chat.addMessage`, gated on `isNewer`) instead of in the handler, so the guard is preserved.

One thing that's still maybe worth keeping: it links the saved (DB-hydrated) message rather than the raw input, so a brand-new chat's tile still builds with its subtitle populated on first paint. That was the reason the call was in the handler.

Side benefit: an out-of-order or backfilled delivery can't regress a chat's preview to an older message, which the unguarded call could.

Pushed as a follow-up.

lib/services/backend/incoming_message_handler.dart
```diff
@@ -379,7 +379,12 @@ class IncomingMessageHandler {
         unawaited(ChatsSvc.setChatHasUnread(c, false, force: true));
       }
 
-      ChatsSvc.updateChat(c, override: true);
+      // The latest message is linked on the guarded sync path (Chat.addMessage,
+      // gated on isNewer), so a freshly-added chat's tile already has its subtitle
+      // on first paint here.
+      if (!ChatsSvc.updateChat(c, override: true)) {
```
zlshames: Instead of using `.updateChat`, use `.findChatByGuid`. One of the `false` exit conditions of `updateChat` is if `headless == true`. So we don't want to end up calling `addChat` if it's headless, which this will. We should conditionally call `updateChat` based on the result of `.findChatByGuid`, or call `addChat`

zlshames (changes requested):

blue-archon:
Switched it. Gated on `findChatByGuid` now instead of `updateChat`'s return, so a headless context (where `updateChat` returns false) can't end up calling `addChat` for a chat that already exists. Pushed.

zlshames (approved):

Issue #2619, opened 2023-12-16
Contacts not showing in Find My friends


cstraw2021:
While all contacts load properly in the iMessage side of Bluebubbles, half my contacts in Find My Friends show up as their emails and not their contact. Even after manually adding the emails to some contacts, the contacts still do not show up properly in Find My.

tneotia:
For the emails that dont show correctly, do you have an iMessage chat with them? And do the contacts show properly on there?

tneotia:
Look within BlueBubbles itself, not in iMessage - does the contact info show for these people in their threads in BlueBubbles?

AndyXheli:
Same issue I have people under Find My and nothing user Bluebubble ![Screenshot_20240130_180150_BlueBubbles.jpg](https://github.com/BlueBubblesApp/bluebubbles-app/assets/59488153/f75c2246-741e-437c-8362-80edec93832a)

tneotia:
Let me know if this is still an issue with the 1.13.0 beta release

AndyXheli:
Hey just tested looks like still nothing for me 

![Screenshot_20240203_103344_BlueBubbles.jpg](https://github.com/BlueBubblesApp/bluebubbles-app/assets/59488153/1ae6ee2e-a4f9-4f46-85d0-b6abcb889550)

![Screenshot_20240203_103341_BlueBubbles.jpg](https://github.com/BlueBubblesApp/bluebubbles-app/assets/59488153/e3f13cfa-76b4-4909-b6c4-e47e57c1b33f)

tneotia:
Sorry, the original message was directed at OP, in your case I am assuming you are on macOS Sonoma, in which case it is expected that FindMy does not work right now.

AndyXheli:
No problem! I'm running 

Server Version: 1.9.4

macOS Version: 14.3.0

AndyXheli:
> Sorry, the original message was directed at OP, in your case I am assuming you are on macOS Sonoma, in which case it is expected that FindMy does not work right now.

Thank you!!

AndyXheli:
Hey @tneotia so I setup a new server running Server Version: 1.9.4  macOS Version: 12.7.3 and installed  version 1.13.0 and still having an issue any suggestions?

tneotia:
@AndyXheli you need the beta server 1.9.5

AndyXheli:
Same thing for me too contacts show as emails and have my profile picture under them not the actual users

[...]

PR #3079, opened 2026-07-08
feat: FindMy map preview on conversation details

DeveloperBlue:
This PR adds FindMy functionality inside the conversation details page, similar to native iOS.

- Adds a FindMy card widget to the conversation details page. The map is bound to any participants that are sharing, plus the user's own location when available.
  If no participants are sharing their location, no card is displayed.
- Tapping the card widget opens a bottomsheet with the fully navigable map, displaying just the conversation participants (and you).
- The card and bottomsheet are styled to match iOS, Material, and Samsung variants.
- Opening a conversation does a light prefetch so the details page already knows whether to reserve space for the card, avoiding layout shift.
- A Conversation settings toggle, **Hide Find My in Conversation Details**, can turn the card off (shown by default).

> I use then when I'm meeting up with friends and need to quickly find what block/building they're at.
<table width="100%">
  <thead>
    <tr>
      <th width="33%">Details</th>
      <th width="33%">Bottomsheet</th>
      <th width="33%">Group Details</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td colspan="3" align="center"><strong>iOS</strong></td>
    </tr>
    <tr>
      <td width="33%" valign="top">Screenshot_20260818_071821_BlueBubbles_Beta</td>
      <td width="33%" valign="top">Screenshot_20260818_071851_BlueBubbles_Beta</td>
      <td width="33%" valign="top">Screenshot_20260818_071911_BlueBubbles_Beta</td>
    </tr>
    <tr>
      <td colspan="3" align="center"><strong>Material / Samsung</strong></td>
    </tr>
    <tr>
      <td width="33%" valign="top">Screenshot_20260818_071934_BlueBubbles_Beta</td>
      <td width="33%" valign="top">Screenshot_20260818_072001_BlueBubbles_Beta</td>
      <td width="33%" valign="top">Screenshot_20260818_072016_BlueBubbles_Beta</td>
    </tr>
  </tbody>
</table>

Most of the design references were taken from my iPad and just googling. Please let me know if you need any changes or have any suggestions. You can also find me as @DeveloperBlue in the discord.

## Developer Notes
1. The largest core change is to `lib/app/layouts/findmy/findmy_controller.dart`.
   I added filtering by contact handles, fit-to-participants bounding, and an optional own-location marker for conversation details. It does not change the behavior on the main FindMy page.
2. If no one in your conversation is sharing their location, there is no card widget.
   On native iOS, there would instead be two buttons "Request Location" and "Share my Location". I don't think BB supports requesting locations just yet.
3. I added a `findmy_handle_matcher` helper that matches FindMy markers to their actual contacts (e.g. iCloud emails to contacts/chat participants). It's not wired up into the main FindMy feature, but it would also partially solve issue #2619.

<details>
<summary>File changes</summary>

`findmy_controller.dart` — participant filtering, camera fit, own-location marker, scoped socket cleanup
`findmy_handle_matcher.dart` — match FindMy friends to chat participant handles
`findmy_participant_prefetch.dart` — session snapshot so the card can be shown/hidden without layout shift
`findmy_map_widget.dart` — preview mode and participant-map camera fitting
`participants_findmy_card.dart` — conversation details map preview card
`participants_findmy_sheet.dart` — expandable navigable map bottomsheet
`conversation_details.dart` — mounts the card when FindMy is supported
`conversation_view.dart` — warms the prefetch when a conversation opens
`settings.dart` / `conversation_panel.dart` / `settings_items_list.dart` — hide-in-conversation-details setting
`findmy_helpers.dart` — finite-coordinate guard for map projection
</details>

### DIFF
--- a/lib/app/layouts/conversation_details/CLAUDE.md
+++ b/lib/app/layouts/conversation_details/CLAUDE.md
@@ -12,6 +12,7 @@ Displayed as a right-side panel on tablet or pushed screen on mobile.
 - `timeframe_picker.dart` — date range selector (for media/docs filtering)
 - `add_participant.dart` — add a member to a group chat
 - `chat_sync_dialog.dart` — progress dialog for re-syncing chat history
+- `participants_findmy_sheet/` — interactive Find My map bottom sheet for chat participants
 
 ## Material 3 Expressive (`material/`)
 The Material/Samsung-only expressive tree — `conversation_details.dart` routes to these instead
@@ -30,6 +31,7 @@ of the iOS widgets below based on `SettingsSvc.settings.skin.value`. See
 - `chat_options.dart` — action buttons (mute, archive, FaceTime, custom avatar, etc.)
 - `participants_list.dart` — group member list
 - `contact_tile.dart` — individual participant row (tappable → contact details)
+- `participants_findmy_card/` — Find My map preview for participants sharing location
 
 **Shared Media**
 - `attachment_section_header.dart` — section label + "Show more" action
--- a/lib/app/layouts/conversation_details/conversation_details.dart
+++ b/lib/app/layouts/conversation_details/conversation_details.dart
@@ -6,11 +6,12 @@ import 'package:bluebubbles/app/layouts/conversation_details/widgets/attachments
 import 'package:bluebubbles/app/layouts/conversation_details/widgets/chat_info.dart';
 import 'package:bluebubbles/app/state/chat_state_scope.dart';
 import 'package:bluebubbles/app/layouts/conversation_details/widgets/chat_options.dart';
+import 'package:bluebubbles/app/layouts/conversation_details/widgets/participants_findmy_card/participants_findmy_card.dart';
+import 'package:bluebubbles/app/layouts/conversation_details/widgets/participants_list.dart';
 import 'package:bluebubbles/app/layouts/conversation_details/widgets/sections/documents/documents_section.dart';
 import 'package:bluebubbles/app/layouts/conversation_details/widgets/sections/links/links_section.dart';
 import 'package:bluebubbles/app/layouts/conversation_details/widgets/sections/locations/locations_section.dart';
 import 'package:bluebubbles/app/layouts/conversation_details/widgets/sections/media/media_grid_section.dart';
-import 'package:bluebubbles/app/layouts/conversation_details/widgets/participants_list.dart';
 import 'package:bluebubbles/helpers/helpers.dart';
 import 'package:bluebubbles/app/layouts/settings/widgets/settings_widgets.dart';
 import 'package:bluebubbles/database/models.dart';
@@ -125,6 +126,7 @@ class _ConversationDetailsState extends State<ConversationDetails> with WidgetsB
                     ? ChatInfo(chat: chat)
                     : ExpressiveChatHeader(chat: chat),
               ),
+              if (ParticipantsFindMyMapCard.isSupported) ParticipantsFindMyMapCard(chat: chat),
               SettingsSvc.settings.skin.value == Skins.iOS
                   ? ParticipantsList(chat: chat)
                   : ExpressiveParticipantsSection(chat: chat),
--- a/lib/app/layouts/conversation_details/dialogs/CLAUDE.md
+++ b/lib/app/layouts/conversation_details/dialogs/CLAUDE.md
@@ -7,6 +7,7 @@ Dialogs shown from within the conversation details / info panel.
 | File | Purpose |
 |------|---------|
 | `add_participant.dart` | Dialog for adding a new participant to a group chat |
+| `participants_findmy_sheet/` | Bottom sheet with interactive map for conversation participants |
 | `address_picker.dart` | Picker for choosing which address/handle to use when sending to a contact with multiple numbers |
 | `change_name.dart` | Dialog for renaming a group conversation |
 | `chat_sync_dialog.dart` | Progress dialog shown during a manual chat-level resync |
--- a/lib/app/layouts/conversation_details/dialogs/participants_findmy_sheet/participants_findmy_sheet.dart
+++ b/lib/app/layouts/conversation_details/dialogs/participants_findmy_sheet/participants_findmy_sheet.dart
@@ -71,10 +71,14 @@ class _ParticipantsFindMyMapSheetBodyState extends State<_ParticipantsFindMyMapS
   void initState() {
     super.initState();
     _mapController = MapController();
+    widget.controller.participantSheetMapController = _mapController;
   }
 
   @override
   void dispose() {
+    if (identical(widget.controller.participantSheetMapController, _mapController)) {
+      widget.controller.participantSheetMapController = null;
+    }
     _mapController.dispose();
     super.dispose();
   }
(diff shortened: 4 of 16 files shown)


DeveloperBlue:
If you could hold off on merging this until #3081 is through, I'll rebase this PR on top of that so it benefits from those bugfixes.

zlshames:
@DeveloperBlue looks good, but can you resolve the conflicts , and I'll complete my review.

Also, if there is no location being shared by any contacts, is the map hidden? I wouldn't want it to show unless there's at least 1 person with location data

zlshames (changes requested):
Also, here are some things to call out as potential issues:

1. `lib/app/layouts/findmy/findmy_controller.dart`: `updateParticipantFilter()` calls fitMapToParticipantMarkers() immediately after changing the filter, but mapController may not be attached yet in the preview card path. We should probably guard this or deferring the fit until the map is ready to avoid no-op / racey camera updates when the card is first built.
2. `lib/app/layouts/findmy/findmy_controller.dart`: `handleForFriendMarker()` creates a new Handle(address: ...) fallback, which can produce a synthetic handle that may not correspond to a real contact and could affect avatar/pop-up rendering. If the matcher can’t resolve a real handle, would it be safer to preserve the original friend handle or explicitly handle the fallback path in the UI?
3. `lib/app/layouts/findmy/widgets/findmy_map_widget.dart`: the friend popup lookup now matches on `controller.friendMarkerKeyFor(e) || e.stableId == markerKey`, but `friendMarkerKeyFor()` can return a random key fallback. That makes popup identity unstable across rebuilds for unresolved friends and could break marker-popup consistency. Would it make sense for it to be a little more stable than a random key?
4. `lib/app/layouts/findmy/findmy_participant_matcher.dart`: this matcher is doing a lot of fuzzy identity work. It would benefit from a few focused tests covering email-vs-phone matching, shared-contact matching, and the “no false positive” cases.
5. `lib/app/layouts/conversation_details/widgets/participants_findmy_card/participants_findmy_card_shared.dart`: the card currently hides itself entirely when `!isLoadingParticipants && visibleParticipants.isEmpty`. Should the empty state show placeholder instead of disappearing?

DeveloperBlue:
Thanks for the review, I'll be circling back to this soon to address those pending issues.

And to your first question: Yes, when no participants are sharing their locations, the map will be hidden/not rendered

DeveloperBlue:
I just finished taking a look at all the issues, and rebased onto the latest changes from `development`.

## Issue Resolutions

#### Issue 1 - Map Fit Timing

- `updateParticipantFilter()` no longer calls `fitMapToParticipantMarkers()` directly. Camera fitting is deferred until the map is actually mounted (via `onParticipantMapReady()`).
- `fitMapToParticipantMarkers()` also guards against early calls in participant mode (e.g. from `getLocations()` or socket updates) by no-oping until the participant map controller is attached and marked ready.

#### Issue 2 - Synthetic Handle Fallback

Resolved by getting rid of the synthetic `Handle` creation entirely.

`handleForFriendMarker()` now resolves the handle in this order: The matched participant `Handle` from the matcher --> `friend.handle` --> `null`

`ContactAvatarWidget` already accepts `Handle?`, so unresolved friends get initials/placeholder instead of a fake contact.

#### Issue 3 - Unstable Marker Keys

Replaced `randomString(6)` with deterministic keys. Primarily uses identity-based keys (`stableId` / `handleAddress` / `title`) that stay constant across location updates. For the rare cases where no identity fields are available, the fallback derives a key from available metadata. 

Marker keys now stay consistent across rebuilds, so popup lookup via `friendMarkerKeyFor()` remains reliable.

#### Issue 4 - Tighter Contact Matching

I completely changed how the `FindMyParticipantMatcher` works to be more deterministic. It originally made a lot of lazy assumptions and loose matches.

BB doesn't really have any areas for tests- if testing is added in the future, this is one area that would benefit and I'd be happy to add it.

#### Issue 5 - Widget Placeholder Fallback

No fallback widget is the preferred solution for now.

Normally on iOS there would be "Share My Location" and "Request Location" buttons in place of the card as fallback, but since there's no way for BB to do that just yet, the next best case is to just show nothing and avoid cluttering the page.

Behavior was also flipped to now be initially hidden by default, so if a conversation has no participants sharing their location, they will never see the card widget.

----

## Other Changes

On top of the requested changes, I also updated the following:

- Opening a conversation does a light prefetch to a new `findmy_participant_prefetch` cache file to check if any participants are sharing. This allows us to decide before the `conversation_details` page is mounted if the map card will be rendered or not, completely eliminating any layout shifting.
- Condensed the individually skinned card and sheet components into just a single card and sheet components.
They were small enough and similar enough that it was worth collapsing and handling skin styles internally.
- Added some additional gating so the card never mounts on unsupported platforms (web, pre-Catalina). This reduces any unnecessary fetches to the FindMy service.
- ~~The sheet map now reuses the same mapController from the card~~ This caused more trouble and complexity than it was worth. The map card and map bottomsheet now have their own map controllers again.


Thanks for catching all the issues, let me know if anything else needs to be adjusted.

zlshames:
I do have one final main concern, and a couple small things... But I promise, this should be it! The issue is with the `socket.off` call:

`FindMyController.onClose()` does:

```dart
// findmy_controller.dart:392
SocketSvc.socket?.off("new-findmy-location");
```

`socket_io_client`'s `Socket.off(eventName)` called *without* a handler argument removes **all** listeners registered under that event name, not just "this instance's" listener. There's no per-caller scoping.

Before this PR that was harmless, because `FindMyController` only had one instantiation site ([`findmy_page.dart:34`](../blob/development/lib/app/layouts/findmy/findmy_page.dart#L34), untagged `Get.put`), so structurally at most one instance existed at a time. Closing it meant clearing the only listener anyway.

This PR adds a second, tagged instantiation path per open conversation:

```dart
// participants_findmy_card.dart:442
_controller = Get.put(FindMyController(participantFilter: _currentParticipants), tag: tag);
```

Since it's tagged, GetX keeps this alongside the main page's untagged instance, so now two (or more) `FindMyController`s can be alive concurrently, each with its own `.on("new-findmy-location", ...)` registration from `onInit()`.

I know it's an edge case, but just trying to cover any future implementations as well.

**Suggested fix**

Scope the removal to this instance's own handler:

```dart
SocketSvc.socket?.off("new-findmy-location", _handleNewFindMyLocation);
```

`socket_io_client` supports the `(event, handler)` overload, and `_handleNewFindMyLocation` is already a stable instance-method tear-off, so this should be a drop-in change with no other side effects.

---

Two smaller things while I'm here, lower priority:

- **Possible use-after-dispose**: `ParticipantsFindMyMapCard.dispose()` unconditionally does `Get.delete<FindMyController>(tag: tag)`, which also disposes `mapController`/`popupController`/timers — but `_openExpandedMap` hands that same controller into the modal bottom sheet and awaits it. If the details page/card is torn down while the sheet is still open (back-nav race, hot reload, etc.), the sheet could end up holding a disposed controller/`MapController`. Might be worth guarding the delete on `!_sheetOpen`, or having the sheet manage its own reference.
- **Performance**: `participantFriendsWithLocation` / `FindMyParticipantMatcher.matches()` can run a synchronous `Handle.findOne()` DB lookup per unmatched friend × participant, called directly from the `Obx` build paths in both the card and sheet. For most chats this is cheap (few participants), but it's recomputed on every `friendsWithLocation` change with no caching. It might be worth memoizing the resolved handle per friend.

DeveloperBlue:
Socket cleanup is now scoped, use-after-dispose edge cases should now be handled, and the participant handle matcher is reworked to be a bit lighter and now has a cache 👍

Please let me know if you find any more issues, I definitely don't mind fixing them. I'd hate to ship anything too buggy :)

DeveloperBlue:
rebased, cleaned up the code a bit, and redesigned the cards for Material m3e

I would also suggest merging #3233 first, and I'll rebase this again with the added benefits of the better contact matching api from that PR

PR #3081, opened 2026-07-09
fix: redacted mode leaks for FindMy and custom chat display names

DeveloperBlue:
This PR addresses two areas where user data slips through redacted mode:

### Custom Chat Display Names
If the user set a custom display name for a chat, redacted mode missed it.
- `redactContactInfo()` did not redact the ``displayName`` properly, only ``title``.
- Some UI used `ChatState.getTitle()`, which returns the raw unredacted title; `ChatState.title` is the correct place to access redacted titles.

`redactContactInfo()` and `unredactContactInfo()` now handle ``displayName``. Relevant places (search, notifications, etc.) were updated to use the redacted titles from `ChatState.title` 

### FindMy
FindMy was leaking Friends names, real map coordinates, and had some inconsistent redacting applied to location statuses.

- Real friend names in the bottomsheet list and popups are now redacted
- Added **findmy_helpers.dart** to handle redacting real coordinates.
It uses a list of preset location coordinates as bases and then applies a hashed jitter. This keeps location markers for the same ID stable and fully anonymous across the app.
- Other information ("Last updated ..." and "Live/Shallow Location" statuses) were redacted in some places but not others.
They are now consistently NOT redacted-- I felt this gives good debug information while not being overly heavy handed with redactions.

FindMy redaction uses the existing "Hide Contact Info" toggle under redacted mode, so no new settings are added.

### DIFF
--- a/lib/app/layouts/conversation_list/pages/search/search_view.dart
+++ b/lib/app/layouts/conversation_list/pages/search/search_view.dart
@@ -366,7 +366,7 @@ class SearchViewState extends State<SearchView> with ThemeHelpers {
                             title: RichText(
                               text: TextSpan(
                                 children: MessageHelper.buildEmojiText(
-                                  chat.getTitle(),
+                                  ChatsSvc.getChatState(chat.guid)?.title.value ?? chat.getTitle(),
                                   context.theme.textTheme.bodyLarge!,
                                 ),
                               ),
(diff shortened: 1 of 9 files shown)


zlshames (approved):

zlshames:
The code looks good. Thank you for the contribution!

PR #3082, opened 2026-07-10
fix: self-heal images left undecodable by an interrupted download

blue-archon:
### Problem

If the app is killed mid-download (e.g. an OOM decoding a large incoming image during boot), the file on disk is left truncated and was never marked `isDownloaded`. The display gate trusts any on-disk file, so the poisoned bytes hit the decoder, the image fails, and it shows "Failed to display image" **permanently** — auto re-download never fires because a file already exists. (Confirmed local-only: the same image rendered fine on another client that downloaded it cleanly.) A second path to the same state: a transfer that completes with 0 bytes gets marked downloaded and trusted forever.

### Fix

- `image_viewer.dart`: treat decode failure as the integrity signal. On the error builder, if the attachment isn't marked `isDownloaded`, re-download once (one-shot guard prevents a loop), then evict Flutter's cached failed `FileImage` decode so the fresh bytes render in place without re-entering the chat. A file that *was* fully downloaded but still won't decode keeps a manual retry button.
- `downloads_service.dart`: don't mark a 0-byte transfer as downloaded — delete it and set the error state so it retries.

Deliberately **not** a `size == totalBytes` check: GIFs are re-encoded client-side and `original=false` can return a converted file, so size legitimately differs from `totalBytes`.

### Note

This replaces the previous error-state "info" button (which opened a dialog showing the decode stacktrace) with a "retry download" button. We can keep the stacktrace affordance alongside it if you'd prefer.

### Testing

Validated over a multi-day testing period (desktop + Android): a poisoned image that previously showed the error forever now re-downloads and renders on the next frame.

### DIFF
--- a/lib/app/layouts/conversation_view/widgets/message/attachment/image_viewer.dart
+++ b/lib/app/layouts/conversation_view/widgets/message/attachment/image_viewer.dart
@@ -37,6 +37,61 @@ class _ImageViewerState extends State<ImageViewer> with AutomaticKeepAliveClient
   PlatformFile get file => widget.file;
   ConversationViewController? get controller => widget.controller;
 
+  // One-shot guard for auto-healing an undecodable image. A process crash
+  // mid-download can leave a truncated file that was never marked downloaded
+  // (isDownloaded != true); existsOnDisk-based gating then trusts the poisoned
+  // bytes and the image fails to decode forever. When that happens we
+  // re-download once. A file that WAS fully downloaded but still won't decode
+  // is left for a manual retry (no auto-loop).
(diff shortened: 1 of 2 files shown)


zlshames:
I've already resolved this issue on the `development` branch, but i think i did so in a better way. The underlying cause was that we downloaded the attachments directly to the final location, so the existence of the file, even an incomplete part of it might trigger some code to think it's "ready".

The solution was to write the file from the http stream, to a different/temp location. Then when it's fully complete, move it to the final location. This would also prevent incomplete downloads from being thought as complete and ready.

PR #3083, opened 2026-07-10
fix: eliminate the background-isolate wedge failure class

blue-archon:
### Context

This is the underlying fix behind #3074. The gap in that thread was to find the root cause rather than just fail-visibly on retries. This is it.

### Root cause

Forensics across three on-device wedges (debug + server logs) converged on one failure class: platform-channel calls made from inside background isolates stop completing around app-pause transitions. The isolate runs all DB work on a single-threaded event loop, so one hung channel call freezes every queued request behind it. Two presentations, both observed:
- warm isolate: a prefs write-through (reply-state persistence, an awaited `SharedPreferences.setString`) hangs and jams the loop;
- fresh isolate: init itself hangs on one of four channel round-trips (`path_provider`, legacy `getInstance`, the legacy->async migration, `SharedPreferencesWithCache.create`), leaving the generation stale with reads and writes piling up unserved (which is why even pure in-memory reads looked "wedged", they never started).

### Fix

- Isolate spawn hands over everything the isolate used to fetch via channels: resolved filesystem paths and package info (a `FilesystemService` handoff), plus a full prefs key/value snapshot that backs an `InMemorySharedPreferencesAsync` platform instance. The whole prefs stack then runs in-memory inside isolates with zero channel calls, all existing call sites unchanged, falling back to the old channel path if the handoff fails. This is why it adds `shared_preferences_platform_interface` as a direct dep. It's layered on top of your current native-readable SharedPreferences backend, so the main engine is untouched.
- Reply-state persistence (pure prefs) and `clearNotificationForChat` (a pure method-channel call, no DB work) no longer round-trip through the isolate at all.

### Recovery + diagnostics

- A watchdog restarts a wedged isolate instead of letting it freeze the app. Init reports per-stage progress so a stalled generation names its exact stage, and cooperative shutdown flushes the isolate log buffer before `Isolate.exit` (wedged generations previously vanished from the log).
- Write-lock escalation: two watchdog fires in quick succession indicate a store-wide DB write deadlock an isolate restart can't clear (only a full app restart can), surfaced with a "needs a restart" notification instead of silently saving nothing.

### Notes

- Primarily an Android background-isolate problem, but the changes are platform agnostic. The desktop branch of the new notification uses `local_notifier` to match the current desktop path.
- Squashed from four commits on a fork. I can split the channel-elimination from the watchdog/recovery if you'd rather review them separately.

### Testing

Validated over a multi-day period on desktop and Android. The three wedge signatures have not recurred since.

### DIFF
--- a/lib/database/database.dart
+++ b/lib/database/database.dart
@@ -312,9 +312,31 @@ class Database {
     }
   }
 
-  /// Wrapper for store.runInTransaction
+  /// Wrapper for store.runInTransaction.
+  ///
+  /// Write transactions are timed: ObjectBox allows one writer per store
+  /// process-wide and blocks the rest natively, so a slow holder (or a
+  /// starved waiter) freezes every other writer in the app. Logging the
+  /// call site of anything that spends >5s in here is the only way to name
+  /// the holder when a store-wide write wedge occurs.
   static R runInTransaction<R>(TxMode mode, R Function() fn) {
-    return store.runInTransaction(mode, fn);
+    if (mode != TxMode.write) {
+      return store.runInTransaction(mode, fn);
+    }
+    final sw = Stopwatch()..start();
+    try {
+      return store.runInTransaction(mode, fn);
+    } finally {
+      sw.stop();
+      if (sw.elapsed > const Duration(seconds: 5)) {
+        Logger.warn(
+          'Write transaction spent ${sw.elapsedMilliseconds}ms (wait + hold) — '
+          'possible write-lock contention. Call site:',
+          trace: StackTrace.current,
+          tag: 'DbTx',
+        );
+      }
+    }
   }
 
   static reset() {
--- a/lib/helpers/backend/startup_tasks.dart
+++ b/lib/helpers/backend/startup_tasks.dart
@@ -64,25 +64,38 @@ class StartupTasks {
     return interopReady;
   }
 
-  static Future<void> _initCoreServices({required bool headless}) async {
+  /// [isolateHandoff]/[onStage]: background isolates receive platform-channel
+  /// values (paths, package info, a prefs snapshot) from the main engine at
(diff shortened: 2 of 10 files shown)


zlshames:
I dont think this is the correct solution for it. I know it's AI generated, and I think the solution it provided is not what I'd want to implement. We need to find a more definitive root cause. I don't think the AI did that.

blue-archon:
Root cause, as concretely as I can put it: the background isolate does all its DB work on one single-threaded event loop, and it makes platform-channel calls (prefs, path_provider) on that same loop. Around an app-pause transition those channel calls stop completing — the reply never comes back. Because it's one loop, the un-returned call sits at the head and every queued DB request behind it is stuck for good. That's the wedge. Two forms I caught on-device:
- warm isolate: an awaited `SharedPreferences.setString` (reply-state write-through) never returns and jams the loop;
- cold isolate: init itself never finishes one of its channel round-trips during spawn (`path_provider` / legacy `getInstance` / the legacy→async migration / `SharedPreferencesWithCache.create`), so the isolate comes up but never serves a single request — which is why even in-memory reads looked wedged; they never got to run.

The actual fix is small: don't make channel calls from the isolate at all. Spawn hands it the resolved paths and a snapshot of prefs, and the prefs stack runs in-memory inside the isolate. No channel call on that loop → nothing to hang on → the failure class is gone rather than retried-around.

Everything else in the PR — the watchdog restart, the write-lock "needs a restart" notification, the per-stage init logging — is recovery/diagnostics, not the fix.

Perhaps we're missing the mark still? But I can say the app has been rock solid with this.

zlshames:
For what it's worth, i made recent updates to SharedPreferencesWithCache. So maybe that resolves this weird handoff issue. But I don't think this was a good solution to the issue to be quite honest with you.

PR #3084, opened 2026-07-10
fix: keep MessagesService alive when the same chat is re-entered via an intent

blue-archon:
### Problem

Re-opening a chat through a share/notification intent can push a second `ConversationView` for a chat that already has one mounted. Its view is still alive (split view keeps it in the pane, or `setAllInactive` cleared `activeChat` while the view persists). Both views resolve the same `MessagesService` via `ensureMessagesSvc(guid)`. When the older route disposes, its `disposeMessagesService` → `close()` flushes the struct and clears the `MessageState`s the newer view is rendering from, so every row falls back to "Error loading message at index N" until the chat is re-entered.

### Fix

Guard at the intent entry point. `IntentsService.openChat` already deduped on `activeChat`, but `activeChat` gets nulled (`setAllInactive`) while the view is still mounted, so the check missed. `ChatsService` now refcounts mounted conversation views, and `openChat` treats the chat as already open when a view for its guid is mounted — not only when it's the active chat. The previous `MessagesService.ownershipEpoch` approach is dropped entirely.

### Testing

The dedupe path is exercised on every intent-open (`openChat` no longer pushes a duplicate when a view for the target guid is mounted).

### DIFF
--- a/lib/app/layouts/conversation_view/pages/conversation_view.dart
+++ b/lib/app/layouts/conversation_view/pages/conversation_view.dart
@@ -65,6 +65,7 @@ class ConversationViewState extends State<ConversationView> with ThemeHelpers<Co
   @override
   void initState() {
     super.initState();
+    ChatsSvc.registerMountedConversation(chat.guid);
     controller.fromChatCreator = widget.fromChatCreator;
     controller.fromSearchResult = widget.initialScrollToGuid != null;
     ChatsSvc.setActiveChat(chat);
@@ -205,6 +206,7 @@ class ConversationViewState extends State<ConversationView> with ThemeHelpers<Co
 
   @override
   void dispose() {
+    ChatsSvc.unregisterMountedConversation(chat.guid);
     routeObserver.unsubscribe(this);
     controller.saveReplyToMessageState(); // P8bda
     super.dispose();
--- a/lib/services/backend/java_dart_interop/intents_service.dart
+++ b/lib/services/backend/java_dart_interop/intents_service.dart
@@ -230,7 +230,14 @@ class IntentsService {
       // comment in init()), so activeChat may be a stale leftover from before the
       // Activity was torn down — always navigate explicitly in that case rather
       // than trusting it to already reflect what's on screen.
-      bool chatIsOpen = !isInitialIntent && ChatsSvc.activeChat?.chat.guid == guid;
+      //
+      // Otherwise the chat counts as open if it is the active chat or a view for
+      // it is still mounted: setAllInactive clears activeChat while the
+      // ConversationView is still on the stack (e.g. split view), and pushing a
+      // duplicate view there races the two over the shared MessagesService.
+      final activeMatch = ChatsSvc.activeChat?.chat.guid == guid;
+      final mountedMatch = ChatsSvc.isConversationMounted(guid);
+      bool chatIsOpen = !isInitialIntent && (activeMatch || mountedMatch);
       Logger.debug("Chat is active: $chatIsOpen", tag: "IntentsService");
 
       setPickedAttachments() {
--- a/lib/services/ui/chat/chats_service.dart
+++ b/lib/services/ui/chat/chats_service.dart
@@ -43,6 +43,15 @@ class ChatsService {
 
   ChatState? activeChat;
 
+  /// How many ConversationViews are currently mounted per chat GUID. Unlike
+  /// [activeChat] (nulled by setAllInactive) and ChatState.isAlive (deadened on
+  /// pause), this stays accurate while the widget is on the navigation stack, so
+  /// an incoming share-sheet / notification intent can tell a chat is already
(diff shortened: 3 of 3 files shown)


zlshames:
Can you explain a flow or chain of events that would produce this issue? I've looked at the fix, and I don't like it. I can't communicate properly what it is I don't like about it, I just think it's a bad resolution to whatever the issue is

blue-archon:
Sure, here's the exact sequence:

1. You're in chat X. Its `ConversationView` has adopted the `MessagesService` tagged with X's guid (`initializeMessagesService` → `ensureMessagesSvc(X.guid)`).
2. A notification for X arrives, or you share something into X from another app. Acting on it pushes a new `ConversationView` for the same chat.
3. The new route runs `initializeMessagesService` on the way in. `ensureMessagesSvc(X.guid)` returns the existing service (same guid → same GetX tag), so the new view attaches to the very instance the old view still holds.
4. Navigator then disposes the old route — incoming route initializes before the outgoing one tears down — so the old view's `disposeMessagesService` runs and calls `_messageService.close()`.
5. That close guts the struct + `MessageState`s the new view is now rendering from. Every visible row falls back to "Error loading message at index N", and stays that way until you back out and re-enter X.

So it's specifically a same-chat re-entry through an intent that pushes a second view onto a shared, tag-keyed service, and the loser of the dispose race guts the winner. A different chat is fine (different tag, different instance); a plain back-out is fine (single owner).

The epoch is just how a disposing view answers "am I still the current owner?" — each adopt bumps a counter on the service and records it on the adopting view; dispose closes only if the counter hasn't moved.

zlshames:
I still dont think this is the proper fix. We track the currently alive chat, so idk why we are going and creating new variables just to prevent the navigation to the new chat. We should just be able to protect against re-navigating if the alive chat matches the one being shared to

blue-archon:
The guard exists, but the share path routes through ChatCreator which intentionally nulls `activeChat` before pushing. Are you wanting to make the re-nav guard key off a live `ConversationViewController` for the target chat instead and drop the epoch?

zlshames:
> The guard exists, but the share path routes through ChatCreator which intentionally nulls `activeChat` before pushing. Are you wanting to make the re-nav guard key off a live `ConversationViewController` for the target chat instead and drop the epoch?

The ChatCreator shouldn't even come into play here. At least you didnt call it out in your steps to reproduce. Intents come into the app via the IntentService or the MethodChannelService. If we need to prevent a new conversation view from being navigated to, we need to guard whatever code is in either of those services to check the active chat before navigating.

blue-archon:
Done — the guard's in `IntentsService.openChat` now, and the epoch is gone.

Regarding keying it off the active chat: `activeChat` alone misses it. `setAllInactive` nulls `activeChat` while the chat's `ConversationView` is still mounted — on a tablet in split view that view stays alive in the pane:

```dart
// ChatsService._setAllInactiveSync
if (clearActive) {
  activeChat?.controller = null;
  activeChat = null;            // the view for this guid can still be mounted
}
```

So an intent for that chat sees no active chat, pushes a second `ConversationView`, and the two race over the shared `MessagesService` (`ensureMessagesSvc(guid)` hands back the same instance) — the older route's dispose then closes it out from under the newer one.

So rather than only comparing against `activeChat`, the guard also checks whether a view for the guid is still mounted:

```dart
// IntentsService.openChat  (was: activeChat check only)
final activeMatch  = ChatsSvc.activeChat?.chat.guid == guid;
final mountedMatch = ChatsSvc.isConversationMounted(guid);
bool chatIsOpen = activeMatch || mountedMatch;
```

`ChatsService` refcounts mounted views (`register`/`unregister` from `ConversationView` init/dispose):

```dart
bool isConversationMounted(String guid) => (_mountedConversationCounts[guid] ?? 0) > 0;
```

No new state on `MessagesService`. I've also reframed the PR description to match.

blue-archon:
Rebased onto 0b42b803d.

They fix opposite failures of the same check, so I kept both:

- `0b42b803d`: on the initial-intent path `activeChat` is a stale positive (Activity torn down, isolate survived underneath) but it would wrongly skip navigation.
- This: on the running path `activeChat` is a false negative, `setAllInactive` clears it while the `ConversationView` is still mounted in a split view (e.g., tablet), a duplicate view gets pushed, and the two race over the shared `MessagesService`, which produces the wall of "Error loading message at index N".

```dart
final activeMatch  = ChatsSvc.activeChat?.chat.guid == guid;
final mountedMatch = ChatsSvc.isConversationMounted(guid);
bool chatIsOpen = !isInitialIntent && (activeMatch || mountedMatch);
```

The `!isInitialIntent` gate covers the mounted check too. The refcount is no more trustworthy than `activeChat` on a fresh tree, so it's skipped on that path as well.

`ChatsService` refcounts mounted conversation views (`register`/`unregister` from `ConversationView` init/dispose). No new state on `MessagesService`.

PR #3090, opened 2026-07-17
fix: stop the log file losing lines to concurrent isolates

blue-archon:
**What**

Every isolate that logs (main, GlobalIsolate, IncrementalSyncIsolate, the Android DartWorker) constructs its own `RotatingFileOutput` against the same file. Two problems fall out of that:

1. `output()` wrote one line per call, so another isolate's line could interleave into the middle of a multi-line event (e.g. a stack trace).
2. Rotation used a per-instance byte counter that only saw that instance's own writes, so instances rotated on a wrong total. Rotation renamed the file, leaving the other isolates' open handles on the rotated-away inode, and everything they wrote after that silently vanished from the collected log.

**Fix**

Write the whole event in a single call so it's atomic against the other isolates (append-mode handle). Decide rotation from the file's real length and truncate in place instead of renaming, so every writer's handle stays valid. Nothing is renamed while handles are open, which also removes the close-before-rename step that existed for Windows (errno 32).

Verified on Linux.

### DIFF
--- a/lib/utils/logger/outputs/rotating_file_output.dart
+++ b/lib/utils/logger/outputs/rotating_file_output.dart
@@ -5,11 +5,26 @@ import 'package:logger/logger.dart';
 import 'package:path/path.dart';
 
 /// A [LogOutput] that writes to [latestFileName] and rotates it once the file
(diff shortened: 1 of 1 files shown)


zlshames (approved):
LGTM

PR #3105, opened 2026-07-21
Cache the SDK/pub in a way that is actually useful + security fixes

jjoelj:


zlshames:
@jjoelj

**1. Unpinned installer scripts now run in the push-to-master context**

The new `warm-cache` job runs `curl -fsSL https://fvm.app/install.sh | bash` on Linux and `choco install fvm -y` on Windows. Neither is pinned to a version or checksum, so it's pulling and executing whatever that endpoint serves at the time. This pattern already existed elsewhere, but those jobs only ever ran with a `pull_request` token (read-only, extra restricted for forks). Now it's running on `push` to master too, i.e. the trusted default token context, on every merge. If `fvm.app` ever got compromised, that's a real problem. Worth pinning the install to a release/commit with a checksum, or switching to `dart pub global activate fvm` instead of piping a shell script.

**2. No explicit `permissions:` block on pr-check.yml**

`desktop-builds.yml` sets `permissions:` explicitly, this file doesn't, so it just inherits the org/repo default token scope. Since this PR is the first time the workflow gets a `push` trigger, seems like a good time to add:
```yaml
permissions:
  contents: read
```

at the top level so the token is least-privilege no matter what the org defaults are.

**3. build-windows/build-linux skipping on push relies on implicit `needs` propagation**

`analyze` has `if: github.event_name != 'push'`, but `build-windows`/`build-linux` don't have their own guard, they just skip because their `needs: analyze` dependency got skipped. Works today, but it's an implicit invariant. If analyze's condition ever changes, these two heavy build jobs could quietly start running full compiles on every master merge. Might be worth adding an explicit `&& github.event_name != 'push'` to their `if:` too, just so it doesn't depend on that side effect.

**4. Actions pinned by tag, not SHA**

Minor and not new to this PR, but `actions/checkout@v6` and `actions/cache@v5` are pinned to mutable tags rather than commit SHAs. Standard supply-chain hardening advice is to pin third-party actions by SHA. Not blocking, just a note for a future cleanup since the file's already being touched.

zlshames (approved):

PR #3132, opened 2026-07-25
fix: harden UnifiedPush delivery against parse failures and oversized payloads

blue-archon:
### Problem

Two issues on the UnifiedPush push path:

1. `UnifiedPushReceiver.onMessage` parsed the payload with a plain `Gson()` and had no try/catch around the handler. Integer fields (ROWIDs, timestamps) could parse as doubles and be mangled, and a single malformed payload threw on a background thread and silently dropped the notification.

2. WorkManager caps a worker's input `Data` at about 10KB. A large push (for example an attachment message) threw when enqueuing the `DartWorker`, so that message was never delivered.

### Fix

1. Parse with `ToNumberPolicy.LONG_OR_DOUBLE` and wrap `onMessage` so a bad payload logs and returns instead of dropping the notification.

2. Spill payloads over 9KB to a cache file and pass a marker path in the worker `Data`; `DartWorker` resolves the marker back to the payload and deletes the file.


blue-archon:
Rebased onto v2.1.0. The only conflict was in DartWorkManager, which #3199 (deliver notification replies when killed) also touched. Resolved so this now sits on top of your `(Boolean)` callback and only adds the oversized-payload spill and the parse hardening.

PR #3135, opened 2026-07-25
fix: harden path handling, Tasker intent receiver, and error-log redaction

cameronaaron:
## Summary

- **Path traversal**: `Attachment.path` interpolated the server-supplied `transferName` directly into a filesystem path. The Windows and Linux/macOS branches stripped path separators; the `default:` branch — Android and iOS, the platforms most users are on — applied no sanitization at all, so a `transferName` of `../../../../shared_prefs/...` would resolve outside the attachments directory. Added `Attachment.sanitizeFileName()`, applied on every platform. Deliberately narrow (only `/`, `.`, and `..` are touched) so legitimate filenames hash to the same path as before and already-downloaded attachments aren't orphaned.
- **Tasker intent receiver** (`ExternalIntentReceiver`): this receiver is exported with no permission, so any app on the device can send it intents.
  - The password comparison used Kotlin's `==`, which short-circuits at the first differing byte — a timing oracle any app could probe by repeating the broadcast. Switched to `MessageDigest.isEqual`.
  - `guidAuthKey` defaults to `""` before setup; the comparison now rejects an empty password/stored-key outright rather than letting `password=""` match.
  - The Tasker reply used an untargeted `sendBroadcast`, putting the server URL on a bus any installed app could listen on — including one whose password guess was rejected. Now `setPackage`-targeted at Tasker.
- Two components were exported with no need to be: `SocketIOForegroundService` and `ForegroundServiceBroadcastReceiver`. Every start/stop site uses an explicit same-app `Intent`, which doesn't require exporting; exporting let any installed app start or stop the socket service. Set `exported="false"` on both.
- **Log redaction side effect**: `ApiInterceptor.onError` stripped `"guid"` and `"password"` before logging a failed request, but did so by calling `.remove()` directly on `err.requestOptions.queryParameters` — the live map on the request — which stripped the auth key off the actual request being reported, not just the log line. Now redacts a copy.

## Test plan

- [x] `flutter analyze`: 302 pre-existing info-level lints, zero errors, zero warnings, before and after
- [x] Verified `sanitizeFileName` preserves ordinary filenames byte-for-byte (including ones containing dots) while neutralizing `..`/`.`/empty and flattening path separators
- [ ] Manual test of the Tasker `GET_SERVER_URL` intent against a real Tasker install
- [ ] Manual test that the foreground service still starts/stops correctly with `exported="false"` on a real device

### DIFF
--- a/lib/database/io/attachment.dart
+++ b/lib/database/io/attachment.dart
@@ -231,15 +231,23 @@ class Attachment {
 
   String get directory => "$baseDirectory/$guid";
 
+  /// SECURITY: flattens a server-supplied [transferName] to one inert path segment,
+  /// so `../../..` can't escape [directory]. Narrow on purpose — legitimate names
+  /// (including `my..file.txt`) are unchanged, so existing files still resolve.
+  static String sanitizeFileName(String name, {String replacement = '_'}) {
+    final flattened = name.replaceAll('/', replacement);
+    if (flattened.isEmpty || flattened == '.' || flattened == '..') return 'attachment';
+    return flattened;
+  }
+
   String get path {
+    final safeName = sanitizeFileName("$transferName");
     switch (Platform.operatingSystem) {
       case "windows":
-        return "$directory/${"$transferName".replaceAll(RegExp(r'[<>:"/\|?*]'), "_")}";
-      case "linux":
-      case "macos":
-        return "$directory/${"$transferName".replaceAll(RegExp(r'/'), "_")}";
+        // `\` is also a separator here, and the rest are illegal in a filename.
+        return "$directory/${safeName.replaceAll(RegExp(r'[<>:"\\|?*]'), "_")}";
       default:
-        return "$directory/$transferName";
+        return "$directory/$safeName";
     }
   }
 
--- a/lib/services/network/http_overrides.dart
+++ b/lib/services/network/http_overrides.dart
@@ -4,6 +4,11 @@ import 'package:universal_io/io.dart';
 
 /// Shared certificate validation logic for both HTTP and WebSocket connections
 /// Returns true if the certificate should be accepted (override bad cert), false otherwise
+///
+/// SECURITY / TODO: this accepts *any* failed certificate (self-signed, expired,
+/// wrong CA) as long as the host matches the configured server, so a MITM on that
+/// host is not detectable. Needs cert pinning or an explicit opt-in setting —
+/// coming back to this. See ENGINEERING-STANDARDS.md §9.6.
 bool shouldAcceptCertificate(X509Certificate cert, String host, int port) {
   String serverUrl = sanitizeServerAddress() ?? "";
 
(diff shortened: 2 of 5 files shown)


lib/database/io/attachment.dart
```diff
@@ -231,15 +231,23 @@ class Attachment {
...
   String get path {
+    final safeName = sanitizeFileName("$transferName");
     switch (Platform.operatingSystem) {
       case "windows":
-        return "$directory/${"$transferName".replaceAll(RegExp(r'[<>:"/\|?*]'), "_")}";
-      case "linux":
-      case "macos":
-        return "$directory/${"$transferName".replaceAll(RegExp(r'/'), "_")}";
+        // `\` is also a separator here, and the rest are illegal in a filename.
+        return "$directory/${safeName.replaceAll(RegExp(r'[<>:"\\|?*]'), "_")}";
```
jjoelj, line 248: This is a raw string. Backslash shouldn't be escaped

lib/database/io/attachment.dart
```diff
@@ -231,15 +231,23 @@ class Attachment {
...
   String get directory => "$baseDirectory/$guid";
 
+  /// SECURITY: flattens a server-supplied [transferName] to one inert path segment,
+  /// so `../../..` can't escape [directory]. Narrow on purpose — legitimate names
+  /// (including `my..file.txt`) are unchanged, so existing files still resolve.
+  static String sanitizeFileName(String name, {String replacement = '_'}) {
+    final flattened = name.replaceAll('/', replacement);
+    if (flattened.isEmpty || flattened == '.' || flattened == '..') return 'attachment';
+    return flattened;
+  }
```
jjoelj, line 241: The worst case scenario before this couldn't have caused any meaningful issue because the resulting path would have been a directory, not a file

lib/services/network/http_overrides.dart
```diff
@@ -4,6 +4,11 @@ import 'package:universal_io/io.dart';
 
 /// Shared certificate validation logic for both HTTP and WebSocket connections
 /// Returns true if the certificate should be accepted (override bad cert), false otherwise
+///
+/// SECURITY / TODO: this accepts *any* failed certificate (self-signed, expired,
+/// wrong CA) as long as the host matches the configured server, so a MITM on that
+/// host is not detectable. Needs cert pinning or an explicit opt-in setting —
+/// coming back to this. See ENGINEERING-STANDARDS.md §9.6.
```
jjoelj, line 11: This is something that can be opened as an issue, not added as a comment...

lib/services/network/http_service.dart
```diff
@@ -368,8 +368,10 @@ class ApiInterceptor extends Interceptor {
 
   @override
   void onError(DioException err, ErrorInterceptorHandler handler) {
-    // Get params without sensitive info
-    final params = err.requestOptions.queryParameters;
+    // Get params without sensitive info. Must be a copy — queryParameters is the
+    // live map on the request, so removing from it in place strips the auth key
+    // off the request itself.
```
jjoelj, line 373: Fairly certain that's the point.

jjoelj (changes requested):
This is unlikely to be merged unless you remove all the no-op changes. AI likes to comment things that should really just be in the PR description or a new issue.

I'll let @zlshames review the Android side

PR #3136, opened 2026-07-25
perf: fix O(n^2) hot paths and remove dead code

cameronaaron:
## Summary

- Removed `lib/helpers/backend/sync/sync_helpers.dart` (218 lines, 95 already commented out) — superseded by `sync_actions.dart`, unreachable from anywhere but its own barrel export, and containing the worst algorithms in the repo (three O(n²) passes per sync batch). Confirmed dead by grepping every call site before removing.
- Real O(n²) / O(n·m) fixes, each replacing a linear scan (or N+1 query) with a precomputed Map/Set index:
  - `ReactionTypes.toList().contains(...)` at 10 call sites (several in the per-message render path) — replaced with a `static const Set`.
  - `MessagesService.loadChunk`'s reaction linker re-scanned the whole sync batch per reaction — indexed by guid.
  - `ChatsService.getFilteredChats` chained up to 7 `.where().toList()` passes inside the conversation list's `Obx` — fused into one pass. Condition order preserved exactly: `&&` short-circuiting decides which `ChatState` fields get read, and reading one is what subscribes the `Obx` to it, so reordering would silently change which chats the list updates for.
  - `CustomGroupFilterChipRow`'s unread-count computation was O(groups × unread × chatsPerGroup) per rebuild — inverted into a guid → group-ids index.
  - `CustomGroupActions` ran one query per guid (leaking each, since it was also never closed) — batched with `guid.oneOf(...)`.
  - `chat_actions.dart`'s chat-latest-message sync mapped chats to IDs then searched back per ID — iterates the chats directly.
  - The v2 handle-ID migration in `database.dart` scanned every handle for every message — on a large synced database, hundreds of millions of comparisons before the app opens. Indexed by id.
  - `attachment_actions.dart` and `chat_options.dart` had the same scan-per-item shape against a batch-sized list; both indexed.
  - `mostRecentSent`/`mostRecent`/`mostRecentReceived` copied and sorted the whole message list to take one element — replaced with a single linear max-scan. `_recomputeDeliveredIndicators` sorted the full outgoing-message list per load/receipt for two scans that never read the order — dropped the sort, added an explicit O(1) tie-break (`_sortsBefore`) to preserve prior behavior on equal timestamps.
  - 12 ObjectBox queries were built and never closed (leaked query handles), including inside a per-guid loop.
  - `ChatMessages.messages` copied the full map to a `List` on every access; hot paths now use the new `messagesView` (lazy, no copy).

## Test plan

- [x] `flutter analyze`: 302 pre-existing info-level lints, zero errors, zero warnings, before and after
- [ ] Manual smoke test of the chat list (filters, custom groups, pinning) and message delivery/read indicators on a device — this PR touches hot paths but the fixes were verified by reading, not by running the app


zlshames:
Do you mind resolving the conflicts so i can review this fully?

PR #3137, opened 2026-07-25
chore: add ENGINEERING-STANDARDS.md and an enforced contract test suite

cameronaaron:
## Depends on #3135 and #3136

This branch includes both PRs' commits so its own contract tests (which assert on those fixes) pass in CI. Please merge #3135 and #3136 first — once they land, rebasing this one will shrink the diff to just the standards doc and test suite.

## Summary

Adds `ENGINEERING-STANDARDS.md` and `test/contracts/` — source-sweep tests that assert the standards rather than just documenting them. They read every `.dart` file under `lib/`, never execute application code, and run in about a second (`dart test test/contracts`, wired into `pr-check.yml`).

Structural rules (query-in-loop, scan-in-loop, sort-in-build) parse a real Dart AST rather than matching text: a regex window around `for (` can't tell a loop body from a callback written inside one, and a contract that fires on correct code gets ignored. Verified in both directions — the suite passes clean, and three deliberately injected violations (an N+1 query, a scan in a loop, an unclosed query) were each caught with the exact delta named, then reverted.

Rules the codebase doesn't satisfy repo-wide yet (`Navigator.of(context)` outside `NavigationService`, raw `Scaffold`/`AppBar`, hardcoded colors) use a ratchet that fails if the count goes up **or** down — up means a new violation, down means a fix landed and the ceiling needs lowering in the same commit, so progress can't silently backslide either way.

## Test plan

- [x] `dart test test/contracts`: 27/27 passing
- [x] `flutter analyze`: 302 pre-existing info-level lints, zero errors, zero warnings
- [x] Verified the gate actually fails: injected an N+1 query, a scan in a loop, and an unclosed query; all three caught with the exact delta named, then reverted


cameronaaron:
@jjoelj This included a path transversal vulnerability while imessage sanitizes these paths it would be best to fix it in the app as well. I think at the very least it should be cherrypicked

jjoelj:
I left that PR open, we'll review it

PR #2711, opened 2024-04-05
Abtract backend stuff

TaeHagen:
New and improved abstractions!

PR #3138, opened 2026-07-25
chore: add BackendService abstraction seam

cameronaaron:
## Summary

Adds `BackendService` — an interface between the app and whatever actually delivers iMessages — plus `BlueBubblesBackend`, the stock implementation that forwards to `HttpSvc`. **Behaviour is unchanged**; this is indirection over calls that were already being made.

This re-opens the idea from #2711, which was closed with *"If you think that you want to re-implement it against `master` (post-rewrite), feel free"* — that PR was written against the pre-rewrite tree (`lib/models/`, `http.*`, `ss.*`) and no longer applies.

### Two rules the interface enforces

**1. Return domain objects, not transport payloads.** The outgoing pipeline carried a raw `Map<String, dynamic>` and called `Message.fromMap(data['data'])` itself — hardcoding the BlueBubbles REST response shape into the send path. `_sendWithRace` / `_finalizeOutgoingSuccess` now carry a `Message`.

**2. Gate features on capabilities, not server facts.** UI asked things like:

```dart
if (SettingsSvc.serverDetails.isMinVentura && SettingsSvc.serverDetails.supportsEditAndUnsend)
```

which encodes *"which macOS/server build is this"* into widget code. It now asks `BackendSvc.canEditUnsend`. The BlueBubbles implementation returns exactly the previous expression, so no gate changes meaning.

`_resolveMethod` (private-api vs apple-script) moved into `BlueBubblesBackend` — the transport choice is a BlueBubbles-server concept and doesn't belong in the shared outgoing handler.

### Migrated
The four send paths, edit/unsend, leave chat, group icon set/delete, iMessage availability, and the edit-unsend / Find My / focus-state / scheduled-message / group-management gates.

### Deliberately left on `HttpSvc`
The interface still declares these so a second backend has the full seam, but the call sites are unchanged — each needs more care than a mechanical swap:

- **chat participant add/remove** — consumes the raw response payload to feed `ChatInterface.bulkSyncChats`
- **markRead/markUnread, typing** — isolate actions keyed by chat GUID, not main-isolate calls
- **attachment downloading** — owned by `AttachmentDownloadService` and its reactive progress model
- **`message_api.dart`, `private_api_panel.dart`** — genuinely BlueBubbles transport and configuration

### Scope of the interface, honestly

**17 of the 35 interface members are wired up today**; the other 18 are declared so a second backend has the whole contract, but nothing in-tree calls them yet (participant ops, typing, markRead/Unread, createChat/renameChat, the account-info trio, and several capability flags). If you'd rather this PR only declare what it actually uses, say so and I'll trim it to the wired subset — the migrated call sites don't depend on the rest.

`sendAttachment` returns a record `({Message message, List<Attachment> attachments})` so the confirmed attachments still reach GUID reconciliation without changing what flows into `_matchMessageWithExisting` (which has explicit `dbAttachments`-preservation logic).

## Why

OpenBubbles is a fork that swaps the macOS server for rustpush. I measured the actual divergence: it's **97% additive** — 106k lines added, only ~3,240 of our lines deleted. A trial merge of current `master` into their branch produces **145 conflicted files / 589 hunks**, and only ~8% of those are genuine backend differences. Most of the rest is mechanical.

This is the structural part: with the seam upstream, replacing the backend becomes "add an implementation" rather than "rewrite call sites." It costs us nothing behaviourally and arguably reads better — `canEditUnsend` beats `isMinVentura && supportsEditAndUnsend` at a widget call site.

Happy to drop or reshape any of it if you'd rather the seam sit somewhere else.

## Tests

The repo has no test suite, and the capability getters are the part of this change most likely to drift silently — a wrong boolean hides or shows a feature with no error anywhere. So this adds one.

`BlueBubblesBackend` takes injectable `serverDetails` / `privateApiEnabled` (defaulting to the live `SettingsService`), which makes the getters — pure functions of those two inputs — testable without GetIt, a database, or a server.

**14 tests** pin observable answers for six real server/macOS profiles (Mojave/v0.1.0 → Sonoma/v1.8.0), the private-API toggle, and the exact version boundaries (205 scheduling, 226 group management, 148 + Ventura edit/unsend, Sonoma link previews). Expected values are written literally rather than recomputed from the implementation, so they assert behaviour instead of restating it.

Verified by mutation — each of these breaks the suite:

| Mutation | Tests failed |
|---|---|
| `canEditUnsend` drops the Ventura floor | 3 |
| `needsClientSideUrlPreview` inverted | 4 |
| `canManageGroupChat` drops the private-API gate | 1 |

`flutter_test` moves into `dev_dependencies` — it was a commented-out block stranded under a top-level `objectbox:` key, so it had only been resolving transitively. One-line `pubspec.lock` change. If you'd prefer to keep the repo test-free, the tests and the pubspec change can be dropped without touching the rest.

## Test plan

- [x] `flutter analyze`: zero errors, zero new warnings (302 pre-existing infos + the pre-existing `.env` asset warning, unchanged from master)
- [x] `flutter test`: 14/14 pass, mutation-verified
- [x] No reformatting of untouched code — `git diff` and `git diff -w` are byte-identical
- [ ] **Not runtime-tested against a real server** — I don't have a BlueBubbles setup. Same caveat as #2554. Worth a smoke test of: send text / reaction / multipart / attachment, edit + undo send, leave group, set + remove group icon

### DIFF
--- a/lib/app/layouts/chat_creator/chat_creator.dart
+++ b/lib/app/layouts/chat_creator/chat_creator.dart
@@ -147,8 +147,8 @@ class ChatCreatorState extends State<ChatCreator> with ThemeHelpers {
   void addSelected(SelectedContact c) async {
     selectedContacts.add(c);
     try {
-      final response = await HttpSvc.handle.handleiMessageState(c.address);
-      c.serviceType.value = response.data["data"]["available"] == true ? ChatServiceType.iMessage : ChatServiceType.sms;
+      final available = await BackendSvc.handleiMessageState(c.address);
+      c.serviceType.value = available == true ? ChatServiceType.iMessage : ChatServiceType.sms;
     } catch (e, s) {
       Logger.warn("Failed to check iMessage availability for contact", error: e, trace: s, tag: 'ChatCreator');
     }
--- a/lib/app/layouts/chat_creator/chat_creator_controller.dart
+++ b/lib/app/layouts/chat_creator/chat_creator_controller.dart
@@ -236,8 +236,7 @@ class ChatCreatorController extends StatefulController {
 
   Future<void> _fetchIMessageState(SelectedContact contact) async {
     try {
-      final response = await HttpSvc.handle.handleiMessageState(contact.address);
-      final available = response.data['data']['available'] as bool?;
+      final available = await BackendSvc.handleiMessageState(contact.address);
       contact.serviceType.value = available == true
           ? ChatServiceType.iMessage
           : available == false
--- a/lib/app/layouts/conversation_details/conversation_details.dart
+++ b/lib/app/layouts/conversation_details/conversation_details.dart
@@ -155,9 +155,7 @@ class _ConversationDetailsState extends State<ConversationDetails> with WidgetsB
                         onAttachmentsLoaded: onAttachmentsLoaded,
                       ),
                     ),
-                    if (chat.handles.length > 2 &&
-                        SettingsSvc.settings.enablePrivateAPI.value &&
-                        SettingsSvc.serverDetails.supportsGroupChatManagement)
+                    if (chat.handles.length > 2 && BackendSvc.canLeaveChat)
                       SliverToBoxAdapter(
                         child: Builder(builder: (context) {
                           return ListTile(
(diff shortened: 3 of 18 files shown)


jjoelj:
> I don't have a BlueBubbles setup.

cameronaaron:
@jjoelj The original PR by @TaeHagen had the same note and was left open for nearly a year

cameronaaron:
But also its not a bad point

jjoelj:
That one wasn't AI slop

PR #3139, opened 2026-07-26
chore: keep deprecated short-form service accessors as aliases

cameronaaron:
## Summary

Adds 11 one-line `@Deprecated` aliases for the pre-rename service accessors (`ss`, `http`, `chats`, `ns`, `ts`, `fs`, `ls`, `mcs`, `notif`, `sync`, `as`), each pointing at its current replacement. Nothing in this repo uses them.

## Why this is worth 11 lines

The rename (`ss` → `SettingsSvc` etc.) deleted the old accessors outright. The cost of that isn't the merge conflicts — it's that **code stops compiling in files git merges cleanly**, which is silent until build time.

Measured against the OpenBubbles fork, the only real downstream consumer:

| | Count |
|---|---|
| Files git reports as conflicted on a full merge | 145 |
| Conflict hunks attributable to the service rename | ~90 |
| **Old-accessor references across their tree** | **3,263** |

So ~90 references announce themselves as conflicts and get fixed. The other **~3,170 are in cleanly-merged files** and just break the build. These aliases cover **3,184 of the 3,263 (97.6%)**.

This isn't only about forks — the same applies to any long-lived branch or out-of-tree patch written before the rename.

## Deliberately not aliased

- **`cs`** — the old `ContactsService` became `ContactServiceV2`, which has a *different API*. An alias would compile while silently changing behaviour; a compile error is the better outcome. (38 refs, intentionally left broken.)
- **`ah`, `cm`, `inq`, `outq`, `fdb`, `upr`** — those services no longer exist in any form, so there's nothing to point at.
- **`setup`** — never renamed; still available under its original name.

## Tests

3 tests assert each alias resolves to the **same instance** as its replacement — a wrong target would still compile wherever two services share a supertype, so identity is the thing worth checking — and that an unregistered service still throws *through* the alias rather than being masked by it.

(The tests run today because `flutter_test` resolves via the SDK. #3138 moves it into `dev_dependencies` properly; whichever of the two lands first can carry that one-line change.)

## Test plan

- [x] `flutter analyze`: **302 issues before and after**, zero errors
- [x] No identifier collisions from exporting names like `as`, `http`, and `sync` through the services barrel — this was the main risk and it's clean
- [x] `flutter test`: 3/3 pass

## Reverting

Delete `lib/services/legacy_accessors.dart` and its one export line. That's the whole footprint — worth saying explicitly, since carrying deprecated shims is a real (if small) maintenance cost and you may reasonably not want it.

### DIFF
--- a/lib/services/legacy_accessors.dart
+++ b/lib/services/legacy_accessors.dart
@@ -1,0 +1,59 @@
+/// Deprecated short-form service accessors.
+///
+/// These were the accessor names before services moved to the `XxxSvc` +
+/// GetIt convention. They are kept as thin aliases so that code written
+/// against the old names — long-lived branches, out-of-tree patches, and
+/// downstream forks — keeps compiling across the rename instead of failing
+/// at every call site.
+///
+/// This matters more than the rename's merge conflicts suggest: a textual
+/// conflict is visible and gets resolved, but a file that merges *cleanly*
+/// and then fails to compile is silent until build time, and there are far
+/// more of those.
+///
+/// Nothing in this repo uses these. They cost one line each, and the file can
+/// be deleted outright once downstream consumers have migrated.
+///
+/// Deliberately **not** aliased:
+/// - `cs` — the old `ContactsService` was replaced by `ContactServiceV2`,
+///   which has a different API. An alias would compile but silently change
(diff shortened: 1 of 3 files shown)


jjoelj:
No

PR #3153, opened 2026-07-31
fix: enable right-click paste for images in text box (#3142)

### LINKED ISSUE #3142: [BUG] cannot right click in text box to paste image
**Description**
Right-clicking in the text box with an image as the most recent clipboard does not result in a menu with a "paste" option. This option does come up when pasting text, and ctrl+v can always be used.

**Expected Behavior**
When right clicking in the message field, a menu with a "paste" option appears.

**Steps to Reproduce**
1. "Copy" an image to the clipboard.
2. Right click in the message field.

**App Version**
2.0.0.0

**Operating System**
Windows 11

Harshit332004:
Closes #3142

Problem:
Right-clicking in the message input field when an image or file is in the clipboard does not display a "Paste" option in the context menu. The default Flutter context menu relies on Clipboard.hasStrings() and ignores non-text media.

Solution:
Modified contextMenuBuilder in text_field_component.dart to intercept the default editableTextState.contextMenuButtonItems.

Checked if a default Paste button is missing from the list.

If missing, implemented a FutureBuilder to asynchronously probe Pasteboard for images or files via a _clipboardContainsMedia() helper.

If media is detected, a custom Paste button is injected into the menu that delegates to the existing clipboardHandler.handlePasteEvent() (mirroring the Ctrl+V behavior).

This approach preserves Flutter's native text paste functionality, gracefully handles exceptions, and is safely gated behind kIsDesktop.

Testing:
Tested locally on Windows. Verified that copying an image to the system clipboard and right-clicking the text input field now successfully renders the "Paste" option, and clicking the option correctly pastes the media into the chat.

### DIFF
--- a/lib/app/layouts/conversation_view/widgets/text_field/text_field_component.dart
+++ b/lib/app/layouts/conversation_view/widgets/text_field/text_field_component.dart
@@ -405,6 +405,19 @@ class TextFieldComponentState extends State<TextFieldComponent> {
                               ],
                             );
 
+                          toolbar.buttonItems?.addAllIf(
+                            !(toolbar.buttonItems?.any((item) => item.label == 'Paste') ?? false),
+                            [
+                              ContextMenuButtonItem(
+                                onPressed: () {
+                                  clipboardHandler?.handlePasteEvent();
+                                  editableTextState.hideToolbar();
+                                },
+                                label: 'Paste',
+                              ),
+                            ],
+                          );
+
                           // Use outerTheme (captured from the real build context) because the
                           // contextMenuBuilder's own `context` parameter shadows the build context
                           // and may not have the dark theme applied (it's a detached overlay context).


jjoelj:
Can you remove the formatting changes or put them in a separate commit please

Harshit332004:
The formatting changes have been completely removed, and I split the import order fix into a separate commit to keep the history clean.

jjoelj:
> The formatting changes have been completely removed

No they have not

Harshit332004:
Yeah, that was my bad. I've pushed a new commit with a completely different approach.

Instead of using a FutureBuilder to probe the clipboard, I now just check if the default "Paste" button is missing from the context menu (which happens when the clipboard has no text, e.g. only an image). If it's missing, I inject one that calls the same clipboardHandler.handlePasteEvent() that Ctrl+V already uses. Much simpler and synchronous.

Tested on Windows 11 — copied an image, right-clicked the text field, "Paste" shows up and correctly attaches the image.

No formatting changes this time, only 13 lines added. Sorry for the mess before!

jjoelj (changes requested):
image

This seems to not work properly if there is text copied to the clipboard. Also, this should work in the subject field if a user has it enabled.

PR #3164, opened 2026-08-04
fix: persist the Firebase config clear before exiting

blue-archon:
"Clear Configurations" in the Firebase panel doesn't stick: after clearing and relaunching,
Firebase shows as "Configured" again and the app re-registers with FCM.

Root cause: deleteFcmData() clears the FCM config from SharedPreferences (SharedPreferencesWithCache
over the async Android store), then the panel calls exit(0) a few ms later. The prefs removal is
written asynchronously, and the hard exit(0) tears the process down before that background write
flushes to disk, so the on-disk config survives and reloads on next launch. (The ObjectBox copy is
cleared durably; it's the prefs mirror that comes back, and getFCM() falls through to it.)

Confirmed with a debug build: immediately after clearConfig() the in-memory read returns cleared,
but the very next launch reads the config back from prefs.

Fix: let the async prefs write flush before exit(0). It's a brief wait before exit.

Repro: Firebase panel -> Clear Configurations -> confirm (app exits) -> reopen. Status reads
"Configured" again.


zlshames (approved):
Simple enough. LGTM

PR #3165, opened 2026-08-04
fix: don't rebuild FCM config from the SharedPreferences mirror

blue-archon:
getFCM() treats the SharedPreferences copy of the FCM config as a fallback: when the ObjectBox
store is empty it rebuilds FCMData from prefs. That mirror can lag behind the database, so an FCM
config that was cleared from ObjectBox gets resurrected from a stale prefs copy.

ObjectBox is the durable source of truth here (FCMData.save() writes both stores together, and the
DB is the primary). When it's empty, return an unconfigured FCMData instead of reading the prefs
mirror. A cleared config then stays cleared. No change for configured installs, since the DB has
the record and is returned as before.

This is independent of #3164 (they address the same "clear doesn't stick" symptom at different
layers) and applies as defense-in-depth even if the prefs write in that PR flushes correctly.

### DIFF
--- a/lib/database/io/fcm_data.dart
+++ b/lib/database/io/fcm_data.dart
@@ -70,17 +70,13 @@ class FCMData {
   }
 
   static FCMData getFCM() {
+    // ObjectBox is the durable source of truth for FCM configuration. When it's empty the
+    // config isn't present, so return an unconfigured FCMData rather than rebuilding it from
+    // the SharedPreferences mirror. That mirror can lag behind the database (e.g. a clear
+    // whose async prefs write didn't flush before the app exited), and reading it here would
+    // otherwise resurrect a configuration the user already cleared.
     final result = Database.fcmData.getAll();
-    if (result.isEmpty) {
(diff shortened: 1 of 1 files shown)


zlshames:
If the firebase config changes, wouldnt that get saved to the shared preferences? If it's not, why wouldnt it be? Should we try to make sure that whenever the fcm data changes, we sync the settings to the shared prefs?

blue-archon:
Good point, and you're right. `FCMData.save()` already writes both stores on every change (`Database.fcmData.put` plus `PrefsSvc.firebase.saveConfig`), and `deleteFcmData()` clears both, so the config does stay mirrored to shared prefs. The stale read wasn't a missing sync. It was the "Clear Configurations" button's async prefs clear not flushing before the panel's `exit(0)` killed the process, which #3164 (now merged) fixes by flushing first.

With #3164 in, the mirror stays consistent, so this change is redundant. Closing it out. Thanks for the review.

PR #3184, opened 2026-08-10
Update startup/splash screen on Desktop

### LINKED ISSUE #3146: Add a close button to the desktop splash screens
- Maybe after a delay, to allow killing app if something goes wrong on startup
- Suggest fixes or to submit Github issue?

jjoelj:
Closes #3146

Adds close button after 3 seconds of startup
Adds button to open Github issue if startup takes too long

### DIFF
--- a/lib/database/database.dart
+++ b/lib/database/database.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:bluebubbles/helpers/backend/startup_tasks.dart';
 import 'package:bluebubbles/helpers/helpers.dart';
 import 'package:bluebubbles/database/models.dart';
 import 'package:bluebubbles/database/migrations/chat_latest_message_migration.dart';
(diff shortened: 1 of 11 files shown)


zlshames:
One thing I'm worried about is users closing the splash screen before a migration finishes. Let's say a migration is in progress and it's taking a long time. If a user closes the splash screen, what will they see? A blank conversation list? I assume it wont cancel the migration, but I just want to make sure i understand what this is solving.

jjoelj:
The migration(s) can now send progress to the splash screen, which will show a progress bar for each one. The splash won't show a close button unless it receives nothing from the app for 30 seconds. Regardless, a close during a migration won't cause issues, but this gives more feedback to the user.

Closing the splash screen exits the app. The user won't see any remaining BlueBubbles window. The intent is to be able to close a hanging instance without task manager.

zlshames (approved):

PR #3187, opened 2026-08-10
Fix contact pins not resizing

### LINKED ISSUE #3111: Contact Pins not resizing.
Image

You can see on the top right of the image that the conversation list is clipping pinned messages. On stable resizing that window has all of the message pins resize with it with out and clipping. 

Currently on Desktop Beta 5 Windows.

jjoelj:
Fix #3111

### DIFF
--- a/lib/app/layouts/conversation_list/pages/cupertino_conversation_list.dart
+++ b/lib/app/layouts/conversation_list/pages/cupertino_conversation_list.dart
@@ -16,7 +16,6 @@ import 'package:flutter/material.dart';
 import 'package:flutter_acrylic/flutter_acrylic.dart';
 import 'package:get/get.dart';
 import 'package:smooth_page_indicator/smooth_page_indicator.dart';
-import 'package:universal_io/io.dart';
 
 class CupertinoConversationList extends StatefulWidget {
   const CupertinoConversationList({super.key, required this.parentController});
@@ -112,6 +111,7 @@ class CupertinoConversationListState extends State<CupertinoConversationList> wi
                       int maxOnPage = rowCount * colCount;
                       PageController _controller = PageController();
                       int _pageCount = (pinCount / maxOnPage).ceil();
+                      bool singleRow = usedRowCount == 1;
 
                       return SliverPadding(
                         // Bottom padding guarantees visible separation from the regular chat list even if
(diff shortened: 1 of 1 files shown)


lib/app/layouts/conversation_list/pages/cupertino_conversation_list.dart
```diff
@@ -195,14 +194,14 @@ class CupertinoConversationListState extends State<CupertinoConversationList> wi
                                       return Padding(
                                         padding: const EdgeInsets.symmetric(horizontal: 10),
                                         child: Column(
-                                          mainAxisAlignment: MainAxisAlignment.center,
+                                          mainAxisAlignment: MainAxisAlignment.start,
```
zlshames, line 197: Any reason you changed this to `start`? Pretty sure I made it `center` to match how iOS handles it.

lib/app/layouts/conversation_list/pages/cupertino_conversation_list.dart
```diff
@@ -195,14 +194,14 @@ class CupertinoConversationListState extends State<CupertinoConversationList> wi
...
                                           children: List.generate(usedRowCount, (rowIndex) {
                                             final int rowStart = rowIndex * colCount;
                                             final List<Chat> rowChats =
                                                 pageChats.skip(rowStart).take(colCount).toList();
-                                            final bool singleRow = usedRowCount == 1;
 
                                             return Row(
+                                              mainAxisSize: MainAxisSize.min,
```
zlshames, line 204: How does this change the layout?

zlshames (changes requested):

lib/app/layouts/conversation_list/pages/cupertino_conversation_list.dart
jjoelj: This makes pins that aren't on the first page of pinned chats top-aligned instead of vertically centered. It doesn't affect the first page so there's no ios counterpart

lib/app/layouts/conversation_list/pages/cupertino_conversation_list.dart
jjoelj: If the available space is wider than the pinned container, this makes the rows not expand to fit. If it isnt set this way, the pins left align instead of being centered. This matters when the screen is wide, since we clamp the max size of pins now.

zlshames (approved):

PR #3188, opened 2026-08-10
Add app version to easier location

### LINKED ISSUE #3141: UI Change: Displaying the Desktop App Version
Making this ticket for adjusting the placement of where the Desktop app version is located, potentially on the right sidebar, either at the top of the scroll, or all the way at the bottom. Currently implementation is Settings > Scroll Down to About Links > About Version, which is a decent amount of clicks to find. Would be nice to have it displayed just by scrolling or one single click.

Below is one of the areas where it could be placed, another would be at the bottom of the left scroll.

Image

jjoelj:
Closes #3141

### DIFF
--- a/lib/helpers/ui/ui_helpers.dart
+++ b/lib/helpers/ui/ui_helpers.dart
@@ -17,7 +17,6 @@ import 'package:gesture_x_detector/gesture_x_detector.dart';
 import 'package:get/get.dart';
 import 'package:get_it/get_it.dart';
 import 'package:image/image.dart' as img;
-import 'package:package_info_plus/package_info_plus.dart';
 import 'package:universal_io/io.dart';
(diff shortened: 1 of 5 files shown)


lib/helpers/ui/ui_helpers.dart
```diff
@@ -959,3 +960,104 @@ int? findChildIndexByKey<T>(List<T> input, Key key, Function(T) getField) {
 
   return index == -1 ? null : index;
 }
+
+void showVersion(BuildContext context) {
```
zlshames: Dialogs should be in the UI page's `dialogs` folder. I don't think it belongs in the UI helpers.

zlshames (changes requested):

zlshames (approved):

PR #3191, opened 2026-08-10
Fix dialog colors ios skin

jjoelj:


zlshames (changes requested):
I know you're patching existing code, but instead of this approach, can we create a dedicated dialog for this in the dialog_helpers.dart file? or use an existing one if we can? We need to centralize all dialogs so we can call a single function that will show properly on iOS vs. Material vs. Samsung.

zlshames (approved):

PR #3199, opened 2026-08-10
fix: deliver notification replies when the app has been killed

gsxdsm:
## Problem

Replying from a notification while the app is killed silently drops the message. The notification renders the reply as if it were sent, but nothing is ever delivered.

Three separate bugs stack up to produce this.

### 1. `OutgoingMessageHandler` is never registered in the DartWorker isolate

`initBackgroundIsolate` registers `IncomingMessageHandler` but not its outgoing counterpart, so the send throws before it reaches the network:

```
Worker with method ReplyChat failed! (error: Bad state: GetIt: Object/factory with
type OutgoingMessageHandler is not registered inside GetIt. Did you forget to register it?)
```

The whole handler ran in **2ms** — it never got as far as an HTTP request.

### 2. The recent-reply cache is written before the send resolves

`_handleReplyChat` wrote the dedupe cache up front, then attempted the send. WorkManager correctly retried after the failure above, but the retry found the cache entry, concluded it was a duplicate, and dropped the message.

### 3. `DartWorker` discards the value passed to `MethodChannel.Result.success()`

Returning `false` from a Dart handler means "retry me" (documented in `MethodChannelService._callHandler`), but returning `false` is still a *successful* method-channel call, so it lands in `success()` rather than `error()`. The value was ignored and the work resumed `Result.success()`, converting every retry request into "done" and permanently dropping the event.

**This one is not specific to replies** — it silently discarded `_retry()` from every handler, including `_handleNewMessage` and `_handleNewServerUrl`.

## Also fixed

A failed send is finalized inside `_finalizeOutgoingFailure`, which persists the message with an error code but still completes the item's completer *normally*. Awaiting the completer therefore said nothing about the outcome. The reply handler now re-reads the queued message and requests a retry when it carries an error, and `DartWorkManager` passes the real terminal state to its callback so a reply that did not send renders as "Not sent" instead of appearing delivered.

## Verification

Reproduced and verified on a physical device (Galaxy Z Fold, One UI 8 / Android 17) by killing the app, receiving a message, and replying from the notification.

Before:

```
Received reply to message from Kotlin
Worker with method ReplyChat failed! (Bad state: GetIt ... OutgoingMessageHandler ...)
Worker result RETRY
  -> retry deduped itself against the prematurely-written cache
  -> reported SUCCEEDED, notification showed the reply as sent, nothing delivered
```

After:

```
11:08:42.847  Received reply to message from Kotlin
11:08:43.008  [POST] /api/v1/message/text
11:08:44.246  Updated recent reply cache
11:08:44.248  Worker with method ReplyChat completed successfully
```

Message delivered. Also confirmed that a genuinely failing send (server misconfigured, request timing out) now surfaces as a failure and retries, rather than reporting success.

### DIFF
--- a/android/app/src/main/kotlin/com/bluebubbles/messaging/services/backend_ui_interop/DartWorkManager.kt
+++ b/android/app/src/main/kotlin/com/bluebubbles/messaging/services/backend_ui_interop/DartWorkManager.kt
@@ -19,7 +19,12 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 
 object DartWorkManager {
-    fun createWorker(context: Context, method: String, arguments: HashMap<String, Any?>, callback: () -> (Unit)) {
+    /// [callback] receives whether the Dart work actually SUCCEEDED. Callers must not
+    /// treat completion as success: WorkManager also reaches a finished state on FAILED
+    /// and CANCELLED, which is exactly what happens when a cold engine boot fails while
+    /// the app is killed. Reporting those as success makes the UI claim work was done
+    /// that never happened (e.g. a notification reply that was never sent).
+    fun createWorker(context: Context, method: String, arguments: HashMap<String, Any?>, callback: (Boolean) -> (Unit)) {
         PersistentLog.d(context, Constants.logTag, "Creating new ${Constants.dartWorkerTag} for method $method")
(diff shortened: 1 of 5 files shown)


zlshames (approved):
overall, i think claude did well. LGTM

PR #3201, opened 2026-08-11
fix: open the keyboard when launching straight into a chat

gsxdsm:
## Problem

With **auto open keyboard** enabled, the keyboard does not appear when the app is launched directly into a conversation:

- from a launcher shortcut
- from an `imessage://` URI
- when re-entering a chat that is already open with the keyboard dismissed

Opening a chat normally from the chat list works, which is why this is easy to miss.

## Cause

The composer *does* take focus in all of those cases — this is not a focus bug. Verified on a physical device (Galaxy Z Fold, One UI 8) with `dumpsys` right after a cold start into a chat:

```
mCurrentFocus=Window{... com.bluebubbles.messaging/.MainActivity}   <- window focused
mServedView=io.flutter.embedding.android.FlutterView                <- IME bound to Flutter
mServedInputConnection=RemoteInputConnectionImpl{connection=
    io.flutter.plugin.editing.InputConnectionAdaptor@...}           <- a field IS focused
mInputShown=false                                                   <- keyboard not shown
```

The engine tears down and re-creates its input connection several times while the app is still starting. A show request that lands in one of those gaps is dropped silently, with nothing reported back to the framework. `StartupTasks` already works around the same engine behaviour on the resume path — see [flutter#52599](https://github.com/flutter/flutter/issues/52599).

Opening a chat from the list is unaffected because the app is already settled by then.

## Fix

Ask for the keyboard explicitly, and keep asking until it is actually on screen.

A single `TextInput.show`, and even a fixed set of retries, still loses the race intermittently — measured **2 of 3** cold starts with three fixed retries. So poll instead, and stop as soon as `MediaQuery.viewInsets.bottom` confirms the keyboard is up (bounded to ~5s). It bails out early if the composer loses focus or an overlay/sub-route takes over, so it can never fight the user or a navigation.

Two callers never rebuild the composer, so its `initState` handling cannot help them:

- `IntentsService.openChat` takes its `chatIsOpen` branch and does not navigate — this is the "re-enter an already-open chat" case
- the app-resume path in `StartupTasks` called `requestFocus()` with no show follow-up when `autoOpenKeyboard` was set, so it had the same latent gap

Both now drive the same routine through a new `ensureComposerKeyboard` hook on `ConversationViewController`, registered by the composer while mounted. The resume path skips it when the subject field was the last focused one, so focus is never yanked off the subject line.

## Verification

Reproduced and verified on device by cold-starting straight into a conversation (`am force-stop` then `am start` with the shortcut's intent) and reading `mInputShown` from `dumpsys input_method`, plus manual testing of the shortcut, URI, already-open-chat, and open-from-list paths.

### DIFF
--- a/lib/app/layouts/conversation_view/widgets/text_field/conversation_text_field.dart
+++ b/lib/app/layouts/conversation_view/widgets/text_field/conversation_text_field.dart
@@ -21,6 +21,7 @@ import 'package:bluebubbles/utils/logger/logger.dart';
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:get/get.dart';
 import 'package:image_picker/image_picker.dart';
 import 'package:path/path.dart' hide context;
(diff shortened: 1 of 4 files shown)


gsxdsm:
Added a follow-up commit: **don't fight the user when they dismiss the auto-opened keyboard.**

The retry loop that forces the keyboard up on launch was re-showing it after the user pressed Back — Back hides the IME but leaves the field focused, so the loop saw the bottom inset return to zero and issued another `TextInput.show`, making the keyboard "keep coming back."

Fix: track whether the keyboard has appeared at least once. Only force the show while it has *never* appeared (the cold-start engine race this method exists for); once it has been up, a later zero-inset means the user dismissed it, so stop instead of re-showing.

gsxdsm:
Closing — the auto-open-on-launch approach caused a keyboard-dismiss regression (keyboard re-raises after Back) that needs a different, root-cause fix. Will reopen a cleaner PR.

PR #3202, opened 2026-08-11
fix: stop flashing an FCM registration error toast on transient startup failures

gsxdsm:
## Problem

On startup, `StartupTasks` fires `FirebaseSvc.registerDevice()` in the background and shows an error toast on the **first** failure:

```dart
FirebaseSvc.registerDevice().catchError((e, s) {
  Logger.warn("Failed to register FCM device on startup!", error: e, trace: s);
  showToast("Failed to register FCM device!", isError: true);   // fires on ANY error
});
```

Because registration races network readiness during startup, a momentary hiccup — server briefly unreachable, FCM token not ready yet, transient auth timeout — throws and flashes the toast. In practice this shows up **intermittently on nearly every launch**, for something the user can't act on and that normally succeeds on its own shortly after.

## Fix

Retry the registration with backoff (3s / 10s / 30s) and only surface the toast if it still fails across the whole sequence. Transient startup failures resolve silently; a genuinely persistent failure is still reported.

No behavior change beyond the toast timing — registration itself is unchanged, still runs in the background, and the manual "register" button in Firebase settings is untouched.

### DIFF
--- a/lib/helpers/backend/startup_tasks.dart
+++ b/lib/helpers/backend/startup_tasks.dart
@@ -383,6 +383,33 @@ class StartupTasks {
     Logger.info("Background isolate services initialization complete");
   }
 
+  /// Registers this device with FCM, retrying transient failures before bothering the
+  /// user. The old code showed a "Failed to register FCM device!" error toast on the
+  /// first failure, so a momentary network hiccup during startup (server briefly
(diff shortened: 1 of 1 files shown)


zlshames:
This isn't the root cause. I've already solved it

PR #3211, opened 2026-08-12
bump flutter_contacts to >=2.3.0 to fix MS Exchange contact sync

star-glider:
Per this issue (https://github.com/QuisApp/flutter_contacts/issues/240), any contacts stored in groups not marked as "visible" are silently dropped in flutter_contacts versions prior to 2.3.1; this impacts users with Office365/Exchange contacts.


zlshames:
Good find! LGTM. I'll make you a custom build to test

PR #3248, opened 2026-08-25
fix: hard-exit on Linux when quitting from the system tray

### LINKED ISSUE #3247: Linux/Wayland: tray "Close App" does not quit the process
## Bug

On Linux (reproduced on KDE Plasma / kwin_wayland), **Close App** in the system tray does not quit the process. The window can already be hidden via close-to-tray, but BlueBubbles stays in System Monitor until it is force-killed.

This is independent of Flatpak vs native: I hit it on the official Linux client **v2.1.1+91** (`bluebubbles-bin` from the AUR, same code as current `master`).

## Environment

- OS: CachyOS (Arch), **Wayland** (`kwin_wayland`)
- Client: BlueBubbles **2.1.1+91**
- Setting: **Close to Tray** enabled (this is also the default)
- Tray: Ayatana StatusNotifierItem (`Close App` menu item is present and clickable)

## Steps to reproduce

1. Run the Linux desktop client with **Settings → Desktop → Close to Tray** on.
2. Close the window with the X button (app hides to the tray; process stays up — expected).
3. Right-click the tray icon → **Close App**.

## Expected

The process exits.

## Actual

The process keeps running (socket, isolates, incremental sync). The only way out is SIGKILL / System Monitor.

Invoking the tray menu over D-Bus (`com.canonical.dbusmenu.Event` on the **Close App** item) produces the same result: click is delivered, process never exits.

## Root cause

Tray **Close App** currently does:

```dart
case 'close_app':
  await windowManager.setPreventClose(false);
  await windowManager.close();
  break;
```

`onWindowClose` is what actually calls `exit(0)` on Linux, and only when `isPreventClose()` is already false:

```dart
void onWindowClose() async {
  if (await windowManager.isPreventClose()) {
    await windowManager.hide();
  } else if (Platform.isLinux) {
    exit(0);
  }
}
```

Two problems on Linux/Wayland after close-to-tray has already hidden the window:

1. `window_manager.close()` does not reliably fire `onWindowClose` for a hidden GTK window, so `exit(0)` never runs.
2. Even when close *is* delivered, `setPreventClose(false)` can race, so `onWindowClose` still sees prevent-close and calls `hide()` again (no-op on an already-hidden window).

The title-bar Linux path already special-cases this with a hard `exit(0)`. Tray **Close App** never takes that path.

## Proposed fix

One line in the tray handler, using the same `exit(0)` the title-bar path already uses:

```dart
case 'close_app':
  if (Platform.isLinux) exit(0);
  await windowManager.setPreventClose(false);
  await windowManager.close();
  break;
```

I rebuilt 2.1.1+91 with this change and confirmed the same D-Bus **Close App** click now kills the process in ~2s (it previously left the process running).

Fix is in #3248.

evman:
Closes #3247

## Summary

Linux tray **Close App** does not quit the process when Close-to-Tray has already hidden the window (especially on Wayland). `windowManager.close()` never reliably reaches the Linux `exit(0)` in `onWindowClose`.

This matches the existing title-bar Linux special case: hard-exit instead of depending on a GTK close event for a hidden window.

## Changes

- Add `quitLinuxDesktopApp()` which `exit(0)`s on Linux.
- Call it from the tray **Close App** handler before `windowManager.close()`.
- Route the existing Linux `onWindowClose` exit through the same helper.

Windows/macOS tray close is unchanged.

## Test plan

- [x] Reproduced on KDE Wayland with v2.1.1+91: D-Bus click on tray **Close App** left the process running.
- [x] After this change, the same click exits the process in ~2s.
- [ ] Linux: Close-to-Tray on → X hides to tray → tray **Close App** quits.
- [ ] Linux: Close-to-Tray off → X still quits (existing `exit(0)` path).
- [ ] Windows tray **Close App** still works (no `exit(0)` on that branch).

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -304,25 +304,9 @@ class DesktopWindowListener extends WindowListener {
     if (await windowManager.isPreventClose()) {
       await windowManager.hide();
     } else if (Platform.isLinux) {
-      quitLinuxDesktopApp();
+      exit(0);
     }
   }
-}
-
-/// Quit the desktop process immediately.
-///
-/// On Linux (especially Wayland), `window_manager.close()` does not reliably
-/// fire [DesktopWindowListener.onWindowClose] once the window is already
-/// hidden by close-to-tray. `setPreventClose(false)` can also race so
-/// [onWindowClose] still sees prevent-close and calls `hide()` again. The
-/// title-bar path already hard-exits on Linux; tray "Close App" must do the
-/// same or the process stays alive until SIGKILL.
-void quitLinuxDesktopApp() {
-  if (!Platform.isLinux) return;
-  try {
-    Logger.info("Exiting desktop process", tag: "Desktop");
-  } catch (_) {}
-  exit(0);
 }
 
 class Main extends StatelessWidget {
@@ -637,9 +621,7 @@ class _HomeState extends State<Home> with WidgetsBindingObserver, TrayListener {
         await windowManager.hide();
         break;
       case 'close_app':
-        if (Platform.isLinux) {
-          quitLinuxDesktopApp();
-        }
+        if (Platform.isLinux) exit(0);
         await windowManager.setPreventClose(false);
         await windowManager.close();
         break;
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -304,9 +304,25 @@ class DesktopWindowListener extends WindowListener {
     if (await windowManager.isPreventClose()) {
       await windowManager.hide();
     } else if (Platform.isLinux) {
-      exit(0);
+      quitLinuxDesktopApp();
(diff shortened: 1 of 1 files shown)


jjoelj (changes requested):
This should be a 1 line fix, not 20.

evman:
Collapsed to one line in the tray handler:

```dart
if (Platform.isLinux) exit(0);
```

`onWindowClose` already hard-exits on Linux for the title-bar path, so the helper was unnecessary. `windowManager.close()` still does not fire that handler once close-to-tray has hidden the window, which is why the tray branch needs its own `exit(0)`.

jjoelj (approved):

PR #3253, opened 2026-08-29
Copy 2fa code to clipboard on Windows.

rgelbjw:
Detect a one-time passcode in an incoming message and write it to the clipboard so it can be pasted without opening the chat.

Hooks IncomingMessageHandler._processNewMessage, where every inbound message passes through regardless of transport.

Detection is based on keywords rather than a bare digit match: a 2fa keyword must be present, candidates are ranked by passcode shape, and 4-digit years are rejected. A plain six-digit regex would clobber the clipboard on order numbers and dollar amounts.  The keywords are English only.

Windows-only, where the clipboard is useful and every message arrives on the main isolate. Off by default, since it silently replaces clipboard contents.

Note that I am not a Dart developer (mostly c#), so most of the code was vibe coded.  My inspection didn't reveal any obvious deficiencies but changes should be checked by someone with more expertise.


jjoelj:
image

A similar feature exists natively on Windows.

I don't think we want automatic clipboard overwrite. At most we'd be ok with a button in a chat to copy OTPs from the most recent received message in a chat, where the smart replies currently are.

rgelbjw:
@jjoelj Windows has this feature?  There is a feature that would automatically extract a 2FA code from a BlueBubbles message and put it on the clipboard?  

The way this PR is implemented is that this feature is off by default.  You would have to turn it on in order for it to even kick in.  What you proposed has merit, but introduces an extra step/click when I am sitting on a web page, waiting for 2FA.  When the message arrives, I can simply press Ctrl+V and continue.

image

jjoelj:
@rgelbjw Windows has the feature that I shared a screenshot of, yes.

I understand the point of your feature. I just disagree with placing items on the clipboard unless they are explicitly copied. What if a user is doing something else and has items on the clipboard when the message comes in?

rgelbjw:
@jjoelj That's why it's disabled by default.  

I've used this feature on the Beeper app for half a decade before its recent demise.  I found it super useful and handy.  I can't think of a time where it interrupted by clipboard flow because when it arrives I am waiting for the 2FA to begin with.  But if this were to happen to me, I'd fetch the overwritten clipboard from the Clipboard history (Win+V).

But I also understand your point about the interruption.

PR #3257, opened 2026-09-04
fix: get web client building, syncing messages, and loading contacts

sammorris81:
## Summary
Bundles fixes from an extended `flutter run -d web-server` debugging pass. Before this, the web client wasn't compiling cleanly against several files, wasn't loading message history, and wasn't showing contact names/avatars.

- **Build/compile**: swap `dart:io` for `package:universal_io/io.dart` in `database.dart`, `startup_tasks.dart`, `incoming_message_handler.dart`, `socket_service.dart`, `user_certificates.dart`, and `lifecycle_service.dart`.
- **ObjectBox web stubs**: give `ToOne`/`ToMany` (`lib/database/html/objectbox.dart`) real in-memory relation storage instead of always-null no-ops — several models (`Attachment.message`, `Message.chat`, `Message.handleRelation`, `Handle.contactsV2`, `ContactV2.handles`) silently broke without it. Also flesh out the web `Chat`/`Handle`/`Message`/`Attachment` models to match their `io/` counterparts (`saveAsync`/`toggle*`, `sort()`, `isNewerThan()`, `displayWidth/Height/displayBox`, contact-aware `displayName`/`initials`, wallpaper fields) so shared action/UI code stops throwing on web.
- **Message loading (root cause of the original bug)**: `SyncInterface.bulkSyncData()` now parses messages directly from the raw map on web instead of hydrating from the local ObjectBox boxes, which are never initialized there (`LateInitializationError` on every call) — this was causing message loading to fail and retry forever.
- **Contacts**: route contact sync through the server-fetch path on web (alongside desktop) in `ContactV2Actions`, since `flutter_contacts` has no web support.
- **Setup/sync completion**: `SyncProgress`'s web branch now actually finishes setup (`finishedSetup`, `ChatsSvc.init`, `StartupTasks.onStartup`, `NetworkTasks.onConnect`) instead of relying on the native full-sync completion flow, which never runs on web. Also await the real server-details fetch before the web minimum-version check in `main.dart`.
- **Query layer**: split `chat_stats_queries.dart` and `search_query_helper.dart` into io/web conditional-import pairs, since both were built entirely on ObjectBox query APIs unavailable on web.
- **Isolates**: run isolate requests (`GlobalIsolate.send`/`broadcast`) directly on the main thread on web, since `dart:isolate` doesn't exist there.
- **Debugging tooling**: add a dev-only JS error forwarder to `web/index.html`, gated behind `kDebugMode`, that ships console errors/warnings to a local collector.

Small fixes swept up along the way: a null-safety fix in a `ChatActions` log line, `as int` cast fixes in two migration files, a guard against concurrent `ChatsService.init()` calls, a `CustomGroupsService` web guard, and an unused import removed from `storage_actions.dart`.

## Test plan
- [x] `flutter run -d web-server` compiles and serves cleanly
- [x] Message history loads in a conversation instead of looping on "Failed to fetch message chunk!"
- [ ] Contact names/avatars in the chat list (tracked separately — root-caused but not yet fixed in this PR; see follow-up)
- [ ] Full regression pass against native platforms (Android/iOS/desktop) to confirm no `io/` behavior changed

## Notes for reviewers
This is debugging/stabilization work for an existing, largely-scaffolded web client rather than new feature work — the io/web conditional-import split pattern and ObjectBox web stubs already existed; this PR fixes gaps in that scaffolding that were blocking the client from actually functioning.


sammorris81:
Please ignore, sorry, I didn't mean to issue a PR

PR #3260, opened 2026-09-06
feat: render server-provided contact photos for Find My friends

brendandebeasi:
The Find My friends payload may carry an `avatar` field (base64-encoded contact photo resolved from the host Contacts database by friend handle). This adds support for it.

- `FindMyFriend` parses and round-trips `avatar`.
- The friend list tile renders the decoded photo (matching the standard 40pt avatar, scaled by avatarScale, decode-cached at 256px), falling back to `ContactAvatarWidget` when the field is absent, empty, or undecodable.
- Redacted mode keeps hiding real photos, same as names and addresses.

Backwards compatible: servers that never send `avatar` behave exactly as before. Companion server change: PnutCN/bluebubbles-server#1 (stacked on PR BlueBubblesApp/BlueBubbles-Server#810).

Verification: flutter analyze shows no new issues (252 pre-existing infos before and after, none in the changed files). No runtime test here, per repo convention the PR check workflow compiles all platforms.


jjoelj:
What's the reason to do this over #3233 ?

PR #3279, opened 2026-09-18
Update flutter to 3.47.4

jjoelj:
Fixes an Impeller crash reported by people on Pixel 10 Android 17


KyleFerrigan:
Device: Pixel 10 Pro Fold
OS: Android 17 QPR2 Beta 5

Built this PR and tested on my device: 
This fixes the instant crash from opening the app that occurs on the current public build. 
App functions as expected with messages sending normally.👍

Looking forward to this being merged to a release.

